PR #1, opened 2017-07-31
RPM packaging and some related fixes

jhrozek:
The first patch ships the profile files using a standard Automake suffix variables. I couldn't figure out how to work with wildcards, but maybe we can fix this later. I'm also not sure if this content belongs to lib/Makefile.am or top-level Makefile.am, but this can be also figured out later.

The next two patches are trivial issues.

The last one is a proposed specfile, which works for me (at least builds and rpmlint doesn't complain much).


rpm/authselect.spec
```diff
@@ -0,0 +1,114 @@
...
+%{_datadir}/authselect/default/winbind+fprint/README
+%{_datadir}/authselect/default/winbind+fprint/nsswitch.conf
+%{_datadir}/authselect/default/winbind+fprint/password-auth
+%{_datadir}/authselect/default/winbind+fprint/system-auth
+%{_datadir}/authselect/default/winbind+fprint/fingerprint-auth
+%{_libdir}/libauthselect.so.*
+%{_datadir}/doc/authselect/COPYING
+%license COPYING
+
+%files devel
```
jhrozek: I wasn't sure if we want a -devel package at all. Perhaps we can avoid one at least in fedora until the API stabilizes to avoid risking someone linking ith authselect-devel and us then breaking the API in future version..

pbrezina: Yes, this is probably a good idea. We will just distribute the tool and omit the devel package for now.

pbrezina:
Thank you, I shuffled the makefiles a little bit and move mkdir -p to $top_srcdir/Makefile.am and moved the profile installation to profiles/Makefile.am. Those changes are pushed in master.

jhrozek:
OK, the new commit enables the conditional packaging.

PR #7, opened 2017-08-03
Few code improvements

pbrezina:
Various code improvements.


profiles/sssd/nsswitch.conf
```diff
@@ -1,6 +1,6 @@
-passwd:    files sss
+passwd:    sss files
 shadow:    files sss
```
jhrozek: You don't have to add sss to shadow at all, because we don't implement the shadow map. basically, look at what nsswitch.conf contains in F-27..

jhrozek: but this can be done later, the order is the most important part

src/lib/authselect_files.c
```diff
@@ -19,15 +19,61 @@
...
+        /* Not a conditional line. */
+        return AUTHSELECT_INCLUDE;
+    }
+
+    if (optional == NULL) {
+        /* No options where specified, condition was not met. */
+        return AUTHSELECT_END;
+    }
+
+    line += 2;
```
jhrozek: Shouldn't you check the line length before doing any pointer arithmetics?

src/lib/authselect_util.c
```diff
@@ -80,19 +80,40 @@ trimline(const char *str, char **_trimmed)
...
 
+    filelen = ftell(file);
+    if (filelen == -1) {
+        ret = errno;
+        goto done;
+    }
+
+    rewind(file);
+
+    buffer = malloc_zero_array(char, filelen + 1);
```
jhrozek: Don't you want to have some kind of upper limit to prevent from absurdly large files from starving all memory? I don't think any files authselect handles will be ever larger than a couple of kilobytes..

src/lib/authselect_check.c
```diff
@@ -26,6 +26,43 @@
 #include "lib/authselect_files.h"
 #include "lib/authselect_private.h"
 
+static bool
+check_content(const char *content, const char *expected)
+{
+    const char *nextline;
+
+    /* Each generated file contains a "generated by" preamble that consists
+     * of two comment lines and one empty line. */
```
jhrozek: Wouldn't it be easier to just have the preamble #defined and strcmp the preamble length? But meh..

profiles/sssd/nsswitch.conf
pbrezina: Done.

src/lib/authselect_files.c
pbrezina: We already know that there are atleast two characters: "??". I added a comment.

src/lib/authselect_util.c
pbrezina: I set it to 4MiB which is more than enough. A smaller size would be also doable but since we make the check only as a precaution I wanted to use a bigger value to be sure we don't reject a valid file.

src/lib/authselect_check.c
pbrezina: Not really, since the preamble contains also a data which is generated from time(NULL). The proper way would be to check the text and also parse the date format, but since it is just a comment with no effect, I believe this is enough. I explained it in the code comment.

pbrezina:
Thank you for the review, I addressed your concerns and I'm going to merge it soon.

pbrezina:
Pushed.

PR #12, opened 2017-08-22
activate: check that directories exist and are writable

pbrezina:
Check that directories exist and are accesible for writing and
entering (wx) before we perform any changes.

### DIFF
--- a/src/lib/authselect_activate.c
+++ b/src/lib/authselect_activate.c
@@ -30,6 +30,49 @@
 #include "lib/authselect_paths.h"
 #include "lib/authselect_util.h"
 
+static bool
+check_directories()
+{
+    const char *dirs[] = {
+        AUTHSELECT_CONFIG_DIR,
+        AUTHSELECT_PAM_DIR,
+        AUTHSELECT_DCONF_DIR,
+        AUTHSELECT_DCONF_DIR "/locks",
+        NULL, /* place for nsswitch.conf parent directory */
+        NULL
+    };
+    errno_t ret;
+    bool bret;
+    int i;
+
+    /* We need to special case since nsswitch.conf is a file not a
+     * directory. But we want to make sure that its parent directory
+     * exists and we can write to it.*/
+    dirs[4] = get_dirname(AUTHSELECT_NSSWITCH_CONF);
+    if (dirs[4] == NULL) {
+        ERROR("Unable to get path to nsswitch.conf parent directory!");
+        return false;
+    }
+
+    bret = true;
+    for (i = 0; dirs[i] != NULL; i++) {
+        ret = check_access(dirs[i], W_OK | X_OK);
+        if (ret == EOK) {
+            continue;
+        } else if (ret == ENOENT) {
+            ERROR("Directory [%s] does not exist, please create it!", dirs[i]);
+        } else if (ret != EOK) {
+            ERROR("Unable to access directory [%s] in [wx] mode!", dirs[i]);
+        }
+
+        bret = false;
+    }
+
+    free((char *)dirs[4]);
+
+    return bret;
+}
+
 static errno_t
 write_generated_files(struct authselect_profile *profile,
                       const char **optional)
@@ -169,6 +212,14 @@ authselect_activate(const char *profile_id,
         ERROR("Unable to find profile [%s] [%d]: %s",
               profile_id, ret, strerror(ret));
         return ret;
+    }
+
+    /* Check that all directories are writable. */
+    is_valid = check_directories();
+    if (is_valid == false) {
+        ERROR("Some directories are not accessible by authselect!");
+        ret = EPERM;
+        goto done;
     }
 
     if (force_override) {
--- a/src/lib/authselect_util.c
+++ b/src/lib/authselect_util.c
@@ -27,6 +27,7 @@
 #include <stdlib.h>
 #include <unistd.h>
 #include <limits.h>
+#include <libgen.h>
 #include <sys/stat.h>
 
 #include "lib/authselect_util.h"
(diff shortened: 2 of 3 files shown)


src/lib/authselect_activate.c
```diff
@@ -30,6 +30,70 @@
 #include "lib/authselect_paths.h"
 #include "lib/authselect_util.h"
 
+static char *get_nsswitch_dir()
+{
+    static const char *filename = AUTHSELECT_NSSWITCH_CONF;
```
jhrozek: Did you consider using `dirname(3)` instead?

src/lib/authselect_activate.c
```diff
@@ -30,6 +30,70 @@
...
+
+    dirs[4] = get_nsswitch_dir();
+    if (dirs[4] == NULL) {
+        ERROR("Unable to get path to nsswitch.conf parent directory!");
+        return false;
+    }
+
+    bret = true;
+    for (i = 0; dirs[i] != NULL; i++) {
+        ret = check_access(dirs[i], W_OK | X_OK);
```
jhrozek: So, why do we need this? In general, I'm not a big fan of testing something before action. I think it is preferable to test the action and fail gracefully. Because the DAC access control (permissions) doesn't have to be the only thing that prevents the code from running, later on MAC (=SELinux, AppArrmor, seccomp in containers, ..) can kick you out as well.

I mean, as long as it's about a nice error message, then fine, but I still think handling EPERM gracefully is a better design.

pbrezina:
On 08/22/2017 02:44 PM, Jakub Hrozek wrote:
> *@jhrozek* commented on this pull request.
>
> ------------------------------------------------------------------------
>
> In src/lib/authselect_activate.c
> <https://github.com/pbrezina/authselect/pull/12#discussion_r134468295>:
>
>> @@ -30,6 +30,70 @@
>  #include "lib/authselect_paths.h"
>  #include "lib/authselect_util.h"
>
> +static char *get_nsswitch_dir()
> +{
> +    static const char *filename = AUTHSELECT_NSSWITCH_CONF;
>
> Did you consider using |dirname(3)| instead?

I did, but I don't like this paragraph in dirname manpage:

"Both dirname() and basename() may modify the contents of path,"

pbrezina:
On 08/22/2017 02:46 PM, Jakub Hrozek wrote:
> *@jhrozek* commented on this pull request.
>
> ------------------------------------------------------------------------
>
> In src/lib/authselect_activate.c
> <https://github.com/pbrezina/authselect/pull/12#discussion_r134468763>:
>
>> +        NULL, /* place for nsswitch direcotry */
> +        NULL
> +    };
> +    errno_t ret;
> +    bool bret;
> +    int i;
> +
> +    dirs[4] = get_nsswitch_dir();
> +    if (dirs[4] == NULL) {
> +        ERROR("Unable to get path to nsswitch.conf parent directory!");
> +        return false;
> +    }
> +
> +    bret = true;
> +    for (i = 0; dirs[i] != NULL; i++) {
> +        ret = check_access(dirs[i], W_OK | X_OK);
>
> So, why do we need this? In general, I'm not a big fan of testing
> something before action. I think it is preferable to test the action and
> fail gracefully. Because the DAC access control (permissions) doesn't
> have to be the only thing that prevents the code from running, later on
> MAC (=SELinux, AppArrmor, seccomp in containers, ..) can kick you out as
> well.
>
> I mean, as long as it's about a nice error message, then fine, but I
> still think handling EPERM gracefully is a better design.

Yes, there are more things to kick us out. Although since we run as root 
we can only fail on missing directory. But since we are performing 
changes in system configuration I want to check it before we change 
anything.

Do you think I should implement some backup/rollback mechanism for failures?

jhrozek:
On Tue, Aug 22, 2017 at 06:06:17AM -0700, Pavel Březina wrote:
> On 08/22/2017 02:46 PM, Jakub Hrozek wrote:
> > *@jhrozek* commented on this pull request.
> >
> > ------------------------------------------------------------------------
> >
> > In src/lib/authselect_activate.c
> > <https://github.com/pbrezina/authselect/pull/12#discussion_r134468763>:
> >
> >> +        NULL, /* place for nsswitch direcotry */
> > +        NULL
> > +    };
> > +    errno_t ret;
> > +    bool bret;
> > +    int i;
> > +
> > +    dirs[4] = get_nsswitch_dir();
> > +    if (dirs[4] == NULL) {
> > +        ERROR("Unable to get path to nsswitch.conf parent directory!");
> > +        return false;
> > +    }
> > +
> > +    bret = true;
> > +    for (i = 0; dirs[i] != NULL; i++) {
> > +        ret = check_access(dirs[i], W_OK | X_OK);
> >
> > So, why do we need this? In general, I'm not a big fan of testing
> > something before action. I think it is preferable to test the action and
> > fail gracefully. Because the DAC access control (permissions) doesn't
> > have to be the only thing that prevents the code from running, later on
> > MAC (=SELinux, AppArrmor, seccomp in containers, ..) can kick you out as
> > well.
> >
> > I mean, as long as it's about a nice error message, then fine, but I
> > still think handling EPERM gracefully is a better design.
> 
> Yes, there are more things to kick us out. Although since we run as root 
> we can only fail on missing directory. But since we are performing 
> changes in system configuration I want to check it before we change 
> anything.

OK, that makes sense.

> 
> Do you think I should implement some backup/rollback mechanism for failures?

Maybe in future...at least perhaps we could save the files before
replacing them so the admin can easily roll back? If we save the files
to tmpfs, then either we remove them when the program exits or the next
machine reboot will get rid of the files.

jhrozek:
On Tue, Aug 22, 2017 at 01:06:14PM +0000, Pavel Březina wrote:
> On 08/22/2017 02:44 PM, Jakub Hrozek wrote:
> > *@jhrozek* commented on this pull request.
> >
> > ------------------------------------------------------------------------
> >
> > In src/lib/authselect_activate.c
> > <https://github.com/pbrezina/authselect/pull/12#discussion_r134468295>:
> >
> >> @@ -30,6 +30,70 @@
> >  #include "lib/authselect_paths.h"
> >  #include "lib/authselect_util.h"
> >
> > +static char *get_nsswitch_dir()
> > +{
> > +    static const char *filename = AUTHSELECT_NSSWITCH_CONF;
> >
> > Did you consider using |dirname(3)| instead?
> 
> I did, but I don't like this paragraph in dirname manpage:
> 
> "Both dirname() and basename() may modify the contents of path,"

..which is why these functions typically operate on duplicate of the
input :)

pbrezina:
I created: https://github.com/pbrezina/authselect/issues/16 to track backup-rollback mechanism.

I also updated the patch. I choose to use `dirname` in the end after all, but created a wrapper around it so a dupped string is always returned instead of dupped/static string mishmash.

pbrezina:
Pushed to master.

jhrozek:
OK, retroactive ACK

Issue #30, opened 2018-01-11
authselect select fails if the dconf directory does not exist


jhrozek:
See:
```
authselect --trace select --debug=2 sssd                                                       
[info] [authselect_activate] Trying to activate profile [sssd]                                                        
[info] [authselect_profile] Looking up profile [sssd]      
[info] [authselect_profile_open] Profile [sssd] is a default profile                                                  
[info] [authselect_profile] Profile [sssd] found at [/usr/share/authselect/default/sssd]                              
[info] [read_textfile_dirfd] Reading file [/usr/share/authselect/default/sssd/README]                                 
[info] [read_textfile_dirfd] Reading file [/usr/share/authselect/default/sssd/system-auth]                            
[info] [read_textfile_dirfd] Reading file [/usr/share/authselect/default/sssd/password-auth]                          
[info] [read_textfile_dirfd] Reading file [/usr/share/authselect/default/sssd/smartcard-auth]                         
[info] [read_textfile_dirfd] Reading file [/usr/share/authselect/default/sssd/fingerprint-auth]                       
[info] [read_textfile_dirfd] Reading file [/usr/share/authselect/default/sssd/postlogin]                              
[info] [read_textfile_dirfd] Reading file [/usr/share/authselect/default/sssd/nsswitch.conf]                          
[info] [read_textfile_dirfd] Reading file [/usr/share/authselect/default/sssd/dconf-db]                               
[info] [read_textfile_dirfd] Reading file [/usr/share/authselect/default/sssd/dconf-locks]                            
[error] [check_directories] Directory [/etc/dconf/db/distro.d] does not exist, please create it!
[error] [check_directories] Directory [/etc/dconf/db/distro.d/locks] does not exist, please create it!
[error] [authselect_activate] Some directories are not accessible by authselect!
[error] [authselect_activate] Unable to activate profile [sssd] [1]: Operation not permitted
Unable to activate profile [1]: Operation not permitted
```

This is a minimal docker container, so the directory is not there, but authselect should either own it,require its owner or not fail.

Issue #29, opened 2018-01-11
unused parameter --debug


jhrozek:
If I run:
```
authselect select --help
```

it talks about a `--debug=INT` option, but that appears to be a noop.

Issue #28, opened 2018-01-10
Move custom profiles to /etc


pbrezina:
Having admnistrator custom profiles under /usr is not very correct since it is the place where only packages should place data. We should move it in /etc or optionally in /usr/local.

PR #33, opened 2018-01-12
Multiple fixes

pbrezina:
Resolves:
https://github.com/pbrezina/authselect/issues/30
https://github.com/pbrezina/authselect/issues/29
https://github.com/pbrezina/authselect/issues/28

### DIFF
--- a/rpm/authselect.spec.in
+++ b/rpm/authselect.spec.in
@@ -120,7 +120,14 @@ rm -f $RPM_BUILD_ROOT/%{_libdir}/libauthselect.so
 %{_bindir}/authselect
 %{_mandir}/man8/authselect.8*
 
-%post libs -p /sbin/ldconfig
+%post libs
+/sbin/ldconfig
+# This directory is already created on systems with gdm, but it is not
+# specifically owned by any package. However, it may be missing for example
+# in docker environments thus we need make sure it exists but we should
+# not own it.
+/bin/mkdir -p /etc/dconf/db/distro.d/locks
+
 %postun libs -p /sbin/ldconfig
 
 %changelog


pbrezina:
@jhrozek @sumit-bose Please, review.

rpm/authselect.spec
```diff
@@ -120,7 +120,14 @@ rm -f $RPM_BUILD_ROOT/%{_libdir}/libauthselect.so
...
 %{_mandir}/man8/authselect.8*
 
-%post libs -p /sbin/ldconfig
+%post libs
+/sbin/ldconfig
+# This directory is already created on systems with gdm, but it is not
+# specifically owned by any package. However, it may be missing for example
+# in docker environments thus we need make sure it exists but we should
+# not own it.
+/bin/make -p /etc/dconf/db/distro.d/locks
```
jhrozek: Umm this sounds wrong, as system admin I really wouldn't like to have a post script create a random directory on my filesystem.

pbrezina: And what would be the correct way?

pbrezina:
I added support for `make rpms` and `make srpm`. However, it require to be configured with `CFLAGS=$CFLAGS -fPIC`, otherwise I get this error:

```
Configured with: ../configure --enable-bootstrap --enable-languages=c,c++,objc,obj-c++,fortran,ada,go,lto --prefix=/usr --mandir=/usr/share/man --infodir=/usr/share/info --with-bugurl=http://bugzilla.redhat.com/
Thread model: posix
gcc version 6.3.1 20161221 (Red Hat 6.3.1-1) (GCC) 
configure:2653: $? = 0 
configure:2642: gcc -V >&5 
gcc: error: unrecognized command line option '-V'
gcc: fatal error: no input files
compilation terminated.
configure:2653: $? = 1 
configure:2642: gcc -qversion >&5 
gcc: error: unrecognized command line option '-qversion'; did you mean '--version'?
gcc: fatal error: no input files
compilation terminated.
configure:2653: $? = 1 
configure:2673: checking whether the C compiler works
configure:2695: gcc -m64 -mtune=generic -fstack-protector-all -Wall -Wextra -Wno-sign-compare -Wshadow -Wunused-variable -Wno-unused-parameter -Wno-error=cpp  -O0 -ggdb3 -Werror -Wp,-U_FORTIFY_SOURCE  -Wl,-z,rel
/usr/bin/ld: /tmp/ccFB6Var.o: relocation R_X86_64_PC32 against undefined symbol `__stack_chk_fail@@GLIBC_2.4' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: Bad value
collect2: error: ld returned 1 exit status
configure:2699: $? = 1 
configure:2737: result: no
configure: failed program was:
| /* confdefs.h */
| #define PACKAGE_NAME "authselect"
| #define PACKAGE_TARNAME "authselect"
| #define PACKAGE_VERSION "0.2"
| #define PACKAGE_STRING "authselect 0.2"
| #define PACKAGE_BUGREPORT ""
| #define PACKAGE_URL ""
| /* end confdefs.h.  */  
| 
| int 
| main ()
| {
| 
|   ;   
|   return 0;
| }
configure:2742: error: in `/dev/shm/authselect/rpmbuild/BUILD/authselect-0.2':
configure:2744: error: C compiler cannot create executables
See `config.log' for more details
```

Do you know how to fix it?

rpm/authselect.spec
jhrozek: I think creating the directory under $RPM_BUILD_ROOT and then owning it..

jhrozek:
First thing I'm seeing is the directory creation, I think you wanted to use `mkdir` not `make`:
```
Updating / installing...                                                                                                                                                                                                                    
   1:authselect-libs-0.2-20180123.1141################################# [ 20%]                                                                                                                                                              
make: *** No rule to make target '/etc/dconf/db/distro.d/locks'.  Stop.                                                                                                                                                                     
# GNU Make 4.2.1
# Built for x86_64-redhat-linux-gnu
# Copyright (C) 1988-2016 Free Software Foundation, Inc.
# License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
```

jhrozek:
OK, but at least I'm not seeing the crash. I need to activate the profile with --force (because what I had at the time is authcondfig profile, not authselect)

jhrozek:
Except for the directory creation and ownership, this PR looks good to me.

rpm/authselect.spec
pbrezina: This does sound wrong as well, because we should not own a directory that is created by other package. Even though the package does not specifically own it. I'm not sure which package creates it.

pbrezina:
@jhrozek  I pushed all the patches except directory creation, that remains in this pull request. You do not want it to create the directory during installation, at the same time I don't think it is alright to own this directory since it is gdm specific thing. Would it be alright then if we just skip generation of this file when the directory is not present?

pbrezina:
Closing this pull request. I will open new one about dconf.

PR #36, opened 2018-02-05
Introduce authconfig compatibility tool

pbrezina:
It translates most of the authconfig call into authselect calls together
with necessary configuration files generation.

It is not intended to be 100% compatible with authconfig. The intention is
to make sure that authentication of systems prepared with scripts containing
authconfig calls will succeed.

Otherwise users should migrate to authselect. I will prepare a document that
will help them with this step.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .project
 .cproject
+.pydevproject
 .settings
 .deps
 .libs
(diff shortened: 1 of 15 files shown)


pbrezina:
I added manual changes that contains a guide on how to migrate to authselect from authconfig.

pbrezina:
Pushed to master.

PR #37, opened 2018-02-06
Add support for pam_access and pam_winbind with krb5_auth

pbrezina:
We missed these option in the original desing. Authconfig supports
this so should we.


pbrezina:
Pushed to master.

PR #38, opened 2018-02-06
Add enable and disable feature api and cli command

pbrezina:
This will enable or disable single feature in currently activated profile.


pbrezina:
Pushed to master.

PR #39, opened 2018-02-12
Remove unnecessary functions

dcantrell:
I was interested to see the project replacing authconfig and find authselect to be a really nice new start.  I was reading through the code and noticed some unnecessary functions.  These patches remove that code and replaces it with simple uses of glibc.  I think I also found some memory leaks in the library code for authselect, but will send a separate pull request for that.

### DIFF
--- a/src/cli/main.c
+++ b/src/cli/main.c
@@ -65,7 +65,7 @@ parse_profile_options(struct cli_cmdline *cmdline,
         return ret;
     }
 
-    features = calloc(cmdline->argc, sizeof(const char *));
+    features = malloc_zero_array(const char *, cmdline->argc);
     if (features == NULL) {
         return ENOMEM;
     }
--- a/src/common/Makefile.am
+++ b/src/common/Makefile.am
@@ -13,6 +13,7 @@ noinst_LTLIBRARIES = \
     $(NULL)
 
 libcommon_la_SOURCES = \
+    malloc.c \
     debug.c \
     utils.c \
     $(NULL)
--- a/src/common/common.h
+++ b/src/common/common.h
@@ -41,6 +41,21 @@ typedef int errno_t;
 #define EOK 0
 #endif
 
+/* Wrappers around malloc and realloc to allocate zero-filled
+ * memory and provide type safety. */
+
+void *_malloc_zero(size_t size);
+void *_realloc_array(void *ptr, size_t elsize, size_t num);
+
+#define malloc_zero(type) \
+    (type *)_malloc_zero(sizeof(type))
+
+#define malloc_zero_array(type, num) \
+    (type *)_malloc_zero(sizeof(type) * (num))
+
+#define realloc_array(ptr, type, num) \
+    (type *)_realloc_array((ptr), sizeof(type), (num))
+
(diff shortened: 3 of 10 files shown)


pbrezina:
Hi, thank you for your contribution. I have some patches in my tree that changes the code quite a lot in order to make it easier to unit test and easier to read an understand. I will push the changes during this week and I will look at your patches properly then.

jhrozek:
btw in general +1 to the changes. After quite a few years of C programming, only today I've learned about `reallocarray` .. :-)

pbrezina:
I pushed the changes. There are quite a lot changes in the code so it is possible (and I hope for it) that the memory leaks that you found are fixed. First of all, thank you again for you contribution, David, I appreciate it.

Those functions that you removed are mostly convenient functions that I use so it resembles talloc allocator that we use in sssd. Therefore things like malloc_zero(size, char *) reads better to me then `sizeof` operator and I would like to keep it. It is however possible to use `calloc` in the macro instead of `malloc` and `memset` since it indeed does the same thing, but please keep the macro.

I did not know about `reallocarray` as well, unfortunately so does not my system. It seem to be a BSD extension. I don't have any manual page for it installed and your code (even applied on previous master) did not compile. If we can get it compiled without installing further packages, I would like to use it (in a macro hiding `sizeof` though).

```
/home/pbrezina/workspace/authselect/src/lib/authselect_dir.c: In function ‘authselect_dir_add_profile’:
/home/pbrezina/workspace/authselect/src/lib/authselect_dir.c:85:15: error: implicit declaration of function ‘reallocarray’ [-Werror=implicit-function-declaration]
     resized = reallocarray(dir->profiles, dir->num_profiles + 2, sizeof(char *));
```

Also `_GNU_SOURCE` is already defined by configure (`AC_GNU_SOURCE`) so it does not need to be defined again directly in Makefile. It is defined for `asprintf` that `format()` actually uses internally and its purpose is to hide the fact that `asprintf` returns number instead of a pointer. Again, it is a convenient function that is easier to work with and it resembles `strdup` (or rather `talloc_asprintf()`. I would like to keep it.

dcantrell:
Thanks for taking a look.  I'll rebase these patches and retain the macros as you discussed.  This is likely just a style difference, but I find that macros used in this capacity actually make the code harder to follow than just calling the functions directly.  But either way, it will work.

For the asprintf() wrapper you had to suppress the int return value and go with NULL (like strdup does), that actually _would_ be perfect as a macro.  I know, I contradicted myself.  :)  The reason I suggest that is the use of ... in asprintf() and your wrapper needing to unravel that with va_start which can all be skipped by just using a macro.  Fewer stack frames.

And yes, reallocarray() is from the BSD world and glibc picked it up.  It's convenient and is available when _GNU_SOURCE is defined (technically __USE_GNU).  What system are you using and I will see if I can reproduce the problem.

Thanks for the feedback and I'll send a new pull request.

pbrezina:
I currently use:

$ cat /etc/fedora-release 
Fedora release 25 (Twenty Five)
$ rpm -q glibc
glibc-2.24-4.fc25.x86_64

PR #40, opened 2018-02-16
Simplification of the memory management macros

dcantrell:
Plus an update to po/POTFILES.in so I could get master building.

I retained the macros as discussed, but have it using glibc's calloc() and reallocarray() instead.  The reallocarray() compiler warning was due to the missing malloc.h inclusion, which is required.  The man page states including stdlib.h and defining _GNU_SOURCE is sufficient, but that doesn't work.  Since this function is new as of glibc-2.26, it's no surprise it's in malloc.h.  It's also safe to assume it will remain there for a while and not become part of the GNU extensions since they deem it useful, but otherwise a unique BSD thing.

While looking at the format() function usage, I noticed that in all but one location the return value of format() is checked against NULL and if it's NULL, an error is displayed (sometimes) and ENOMEM returned.  In the instance where this doesn't happen, NULL is returned.  Since you are checking the return value and largely taking the same action everywhere, I think a macro would be sufficient and have it make use of asprintf() via __VA_ARGS__.  I haven't changed anything with format() in this PR, but wanted to work on that separately.


pbrezina:
Thank you. Can you please use `reallocarray` only if it is available and `realloc` otherwise? Configure check can look like this:

```
--- a/configure.ac
+++ b/configure.ac
@@ -33,6 +33,9 @@ m4_include(external/a2x.m4)
 dnl Required libraries
 REQUIRE_POPT

+dnl Check if functions are present
+AC_CHECK_FUNCS_ONCE([reallocarray])
+
 dnl Optional build dependencies - man pages generation
 CHECK_ASCIIDOC_TOOLS
 AM_CONDITIONAL([HAVE_MANPAGES], [test "x$A2X" != "x"])
```

pbrezina:
I pushed the first commit together with my patches so authselect rpm can be built.

dcantrell:
Sure thing, I can do that.  I'll update the PR once I'm in the office on Monday.  I don't have my work laptop with me and am busy this weekend anyway.

pbrezina:
David, I took the liberty and finished this task in https://github.com/pbrezina/authselect/pull/157. I slightly changed the commit message of your commits but otherwise kept them intact to keep your contribution. I added to more commits to finish the usage of `reallocarray`. Feel free to review the PR if you have any spare  time.

PR #41, opened 2018-02-19
MAN: Enhancements to the authselect-migration manual page

jhrozek:
Minor fixes and enhancements to the manual page.


jhrozek:
the group typo was fixed

PR #42, opened 2018-02-19
Fix minor authcompat issues

jhrozek:
There is a number of issues fixed in this PR. First, some pep8 issues were fixed, not all, because I think the long lines and the indent you choose reads better than stricly following pep8 style.

Next, a typo in the sssd.conf template was fixed, one deprecated option was moved and an option help text was amended.


pbrezina:
Thank you. Pushed.

PR #50, opened 2018-04-25
Don't disable oddjobd.service

tiran:
authselect disables oddjobd.service unless the mkhomedir option is
supplied. This breaks other services that depend on oddjobd, e.g.
FreeIPA.

Enable oddjobd.service with mkhomedir. Don't touch oddjobd.service
otherwise.

See: https://pagure.io/freeipa/issue/7465
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1571844
Signed-off-by: Christian Heimes <cheimes@redhat.com>

### DIFF
--- a/src/compat/authcompat.py.in.in
+++ b/src/compat/authcompat.py.in.in
@@ -437,10 +437,10 @@ class AuthCompat:
                 config.cleanup()
                 self.disableService(config.service)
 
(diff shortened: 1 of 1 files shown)


pbrezina:
Thank you. Pushed to master.

PR #59, opened 2018-06-25
write generated files to temporary files first, then rename them

pbrezina:
To avoid breaking the system configuration if we are unable to write
generated content to disc, for example were there is no space left.

### DIFF
--- a/src/lib/files/system.c
+++ b/src/lib/files/system.c
@@ -19,7 +19,9 @@
 */
 
 #include <errno.h>
+#include <stdio.h>
 #include <stdlib.h>
+#include <unistd.h>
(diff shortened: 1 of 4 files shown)


jhrozek:
Two questions:
1. shouldn't the template be created using mkstemp?
2. if renaming fails, don't we leave the templates behind?

pbrezina:
It was much simpler to write it without mkstemp, but yes, it is a better solution. I addressed both your questions in new patch.

pbrezina:
Pushed to master.
* 2f7d4792e0b3b7dd73109f053a60c40293f60d49

Issue #64, opened 2018-07-12
Add systemd nss module to passwd and groups


pbrezina:
This module is present in default Fedora nsswitch.conf and we should include it in our profiles as well.

PR #68, opened 2018-07-13
profiles: add systemd module to nsswitch

pbrezina:
This module is needed to resolve dynamic systemd users and groups. It is
automatically enabled when installing systemd and we should have it enabled
in authselect profiles.

Resolves:
 https://github.com/pbrezina/authselect/issues/64

### DIFF
--- a/profiles/sssd/nsswitch.conf
+++ b/profiles/sssd/nsswitch.conf
@@ -1,5 +1,5 @@
-passwd:     sss files
-group:      sss files
+passwd:     sss files systemd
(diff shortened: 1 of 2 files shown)


jhrozek:
So did the systemd developers confirm that they do want the module in? (FWIW I think it's a good idea to add the module)

pbrezina:
Yes, they did.

pbrezina:
Pushed to master.
* f7aee4dcf2163dacccb74ccf4314fa3cbec11c52

PR #73, opened 2018-08-03
profiles: add support for pam_u2f

SkewedZeppelin:
Adds support for `with-u2f` and `with-u2f-2fa` feature flags for enabling pam_u2f.so in system-auth and password-auth.

Requires pam-u2f package and optionally pamu2fcfg.
Documentation: https://developers.yubico.com/pam-u2f/

Tested working: lightdm, gdm, sudo, gnome screen locker, mate screen locker, polkit prompts
Needs testing: lxdm, sddm, other screen lockers

SELinux policy will need to be updated to allow `login` to read `~/.config/Yubico/u2f_keys`.
```
avc:  denied  { read } for  pid=x comm="login" name="u2f_keys" dev=x ino=x scontext=system_u:system_r:local_login_t:s0-s0:c0.c1023 tcontext=unconfined_u:object_r:config_home_t:s0 tclass=file permissive=0
```

The following will need to be added to the pam-u2f specfile
```
%post
echo "You can now enable the with-pam-u2f and with-pam-u2f-2fa features in authselect"
echo "See /usr/share/authselect/default/*/README for more information"
echo "Warning: pam-u2f does NOT work over SSH and similar"

%postun
authselect disable-feature with-pam-u2f
authselect disable-feature with-pam-u2f-2fa
```

### DIFF
--- a/profiles/nis/README
+++ b/profiles/nis/README
@@ -27,6 +27,12 @@ with-ecryptfs::
 with-fingerprint::
     Enable authentication with fingerprint reader through *pam_fprintd*.
 
+with-pam-u2f::
+    Enable authentication via u2f dongle through *pam_u2f*.
+
+with-pam-u2f-2fa::
+    Enable 2nd factor authentication via u2f dongle through *pam_u2f*.
+
 with-silent-lastlog::
     Do not produce pam_lastlog message during login.
 
--- a/profiles/nis/REQUIREMENTS
+++ b/profiles/nis/REQUIREMENTS
@@ -1,6 +1,12 @@
 Make sure that NIS service is configured and enabled. See NIS documentation for more information.
                                                                                           {include if "with-fingerprint"}
 - with-fingerprint is selected, make sure fprintd service is configured and enabled       {include if "with-fingerprint"}
+                                                                                          {include if "with-pam-u2f"}
+- with-pam-u2f is selected, make sure that the pam u2f module is installed                {include if "with-pam-u2f"}
+  - users can then configure keys using the pamu2fcfg tool                                {include if "with-pam-u2f"}
+                                                                                          {include if "with-pam-u2f-2fa"}
+- with-pam-u2f-2fa is selected, make sure that the pam u2f module is installed            {include if "with-pam-u2f-2fa"}
+  - users can then configure keys using the pamu2fcfg tool                                {include if "with-pam-u2f-2fa"}
                                                                                           {include if "with-mkhomedir"}
 - with-mkhomedir is selected, make sure oddjobd service is enabled                        {include if "with-mkhomedir"}
   - systemctl enable oddjobd.service                                                      {include if "with-mkhomedir"}
--- a/profiles/nis/password-auth
+++ b/profiles/nis/password-auth
@@ -1,6 +1,8 @@
 auth        required                                     pam_env.so
 auth        required                                     pam_faildelay.so delay=2000000
 auth        required                                     pam_faillock.so preauth silent deny=4 unlock_time=1200   {include if "with-faillock"}
+auth        sufficient                                   pam_u2f.so cue                                           {include if "with-pam-u2f"}
+auth        required                                     pam_u2f.so cue nouserok                                  {include if "with-pam-u2f-2fa"}
 auth        sufficient                                   pam_unix.so {if not "without-nullok":nullok} try_first_pass
 auth        requisite                                    pam_succeed_if.so uid >= 1000 quiet_success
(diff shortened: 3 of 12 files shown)


pbrezina:
Thank you for your contribution. I will look into this after 1.0 release.

pbrezina:
One thing that is not clear from the documentation: does this module only read password from the dongle or does it actually perform any kind of authentication?

SkewedZeppelin:
There is no password, the "server" (the pam module) checks if the user has any keys configured, if one is currently connected it sends it a challenge, the user taps the key, the key then provides a response, which gets verified.

This post has a somewhat decent explanation: https://fastmail.blog/2016/07/23/how-u2f-security-keys-work/ and diagram

pbrezina:
Profiles now contains also `REQUIREMENTS` file that can show user additional requirements after selecting the profile. Is there anything that should be suggested to user after enabling with-u2f[-2fa]?

I finally got some yubikey so I can test this.

SkewedZeppelin:
I have rebased and updated `REQUIREMENTS`.

A pull request will still need to be made to https://src.fedoraproject.org/rpms/pam-u2f for adding `%post` to the specfile.

pbrezina:
Package pamu2fcfg needs to be also installed for pamu2fcfg scripts, however I would not use any specific package names in requirements since in general, authselect can be used on any platform not only fedora.

Can you help me with testing this? I installed pam-u2f and pamu2fcfg, then:
```
authselect select sssd with-u2f
pamu2fcfg -u pbrezina > /etc/u2f_mappings

su pbrezina
Password: 
su: Authentication failure
```
I touched the dongle after Password prompt.

SkewedZeppelin:
Try the default location
```
mkdir ~/.config/Yubico
pamu2fcfg > /.config/Yubico/u2f_keys
```
you can also append the `debug` option to the pam settings and get lots of useful output (check if it is checking the correct mapping file, etc), it prints straight to where you are authenticating.


As for the packages:
fedora uses pam-u2f and pamu2fcfg
debian and ubuntu use libpam-u2f and pamu2fcfg
opensuse uses pam_u2f and pamu2cfg
I will update the wording to be more generic ("the u2f pam module")

pbrezina:
I am sorry that this took so long, unfortunately this pull request hit me in a time when I had so many pressing issues. It currently conflicts with master but these are simple to resolve.

I tested the patch and it seems to work correctly. If user has an entry in u2f config it can authentication via dongle, otherwise it goes through other pam modules.

However, I tested only with-u2f, can you please  give me hints how to test with-u2f-2fa?

SkewedZeppelin:
No worries.
I have rebased and added it to NIS as well.

I will have to test again, but iirc:
with-u2f: should try to use key if available, otherwise show password prompt
with-u2f-2fa: will only proceed to password prompt if valid key is available and was invoked by user

pbrezina:
Thank you. I tried both with-u2f and with-u2f-2fa and both seems to work correctly. It works even with gnome login screen which is sweet.

I will ask you for one more change, then I will merge it. Could you please rename the features to `with-pam-u2f`, `with-pam-u2f-2fa`? There may be some integrated solution within sssd and FreeIPA one day an this name would possibly collide. It will also make clear that this is only related to one pam module. Thank you.

SkewedZeppelin:
I have renamed the features.

The only thing left now is that denial (which prevents use on a plain tty) and updating the pam-u2f specfile. I will make the PR for the specfile, but do you have any advice on how to go about getting the denial addressed?

There is already an existing se boolean called `authlogin_yubikey`, but I tested and it seemed to have no effect.

pbrezina:
> The only thing left now is that denial (which prevents use on a plain tty) and updating the pam-u2f specfile. I will make the PR for the specfile, but do you have any advice on how to go about getting the denial addressed?

I do not know what do you mean by "that denial".

SkewedZeppelin:
@pbrezina the one I posted in the beginning at the top above

```
avc:  denied  { read } for  pid=x comm="login" name="u2f_keys" dev=x ino=x scontext=system_u:system_r:local_login_t:s0-s0:c0.c1023 tcontext=unconfined_u:object_r:config_home_t:s0 tclass=file permissive=0
```

pbrezina:
Right. I did not see this denial myself, but the general rule on Fedora is to report it as a bug to selinux-policy package if it is something that should be enabled by default.

pbrezina:
Thank you again for the contribution. I plan to release it as part of  authselect-1.1 (probably Fedora 30), if you would like it to be included sooner please tell me.

Pushed to master:
* b6e227b434124e183a52fcbef206dcf71c2234b5

PR #78, opened 2018-08-28
Don't write options without value to pwquality conf (#1618865)

AdamWill:
Per https://bugzilla.redhat.com/show_bug.cgi?id=1618865 , it is
incorrect to write lines like this in a pwquality config file:

minlen=
minclass=
maxrepeat=
maxclassrepeat=

There should either be an actual integer value, or the line
should be omitted entirely. Including the option with no value
is wrong and breaks pwquality. This should fix the problem by
only writing the lines if the option is actually set.

Signed-off-by: Adam Williamson <awilliam@redhat.com>

### DIFF
--- a/src/compat/authcompat.py.in.in
+++ b/src/compat/authcompat.py.in.in
@@ -319,10 +319,13 @@ class Configuration:
         def write(self):
             config = EnvironmentFile(Path.System('pwquality.conf'))
 
-            config.set("minlen", self.get("passminlen"))
-            config.set("minclass", self.get("passminclass"))
-            config.set("maxrepeat", self.get("passmaxrepeat"))
-            config.set("maxclassrepeat", self.get("passmaxclassrepeat"))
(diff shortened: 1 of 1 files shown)


AdamWill:
Note: from testing, I believe this does fix the bug, but only for fresh installs. If anyone happened to get Rawhide or F29 installed with the bad authselect, then on update to a fixed version, the config file is not regenerated, I think because `authselect apply-changes` considers the existing config valid. I'm not sure of the best way to make it consider a config with any of these "bad" lines in it invalid, and regenerate it?

AdamWill:
I think I also tested with `authconfig --update` and that only worked if the file was wiped (it didn't fix the file if it existed), but I'll check that again. Should the package have `authconfig --update` in %post, or something?

pbrezina:
I'm closing this PR as I opened a new PR that builds up on your patch: https://github.com/pbrezina/authselect/pull/79

PR #79, opened 2018-09-04
compat: write only values set on command line to pwquality.conf

pbrezina:
This is a successor of https://github.com/pbrezina/authselect/pull/78

1. Integer options are never written without a value
2. All options are written only if set on command line, therefore if the option is unset it will not overwrite current value (even if `authconfig --update` is called)
3. Regular expression for parsing values from environment-type config files is fixed so it accepts `-1` as value
4. When updating to this version, rpm scriptlet will fixed the old invalid pwquality.conf


pbrezina:
@AdamWill could you review it please?

pbrezina:
Pushed to master.
* 748e7672134545c5c66fab180c983edbd8d7a106
* 65c081520d2192602ecffb31a06f96e577ed3998
* 7c37960dbf378dc1041c1d635629d509c48bae20
* 81de5acd2ce27592b724ab564ed3ade6ed3d68eb

Issue #82, opened 2018-09-07
compat: use current authselect profile if sysconfig does not exist


pbrezina:
If `/etc/sysconfig/authconfig` does not exist, system is already configured with `authselect` and authselect profile cannot be determined from given `authconfig` options we should not change current authselect profile.

https://bugzilla.redhat.com/show_bug.cgi?id=1626067

PR #84, opened 2018-09-11
compat: use current configuration unless other profile is selected

pbrezina:
This makes sure that 'authconfig --updateall' or 'authconfig --enablexyz --updateall'
will not override current authselect profile if /etc/authconfig/sysconfig does not
exist.

Resolves:
https://github.com/pbrezina/authselect/issues/82


pbrezina:
Users has confirmed that these patches solves the issue.

Issue #100, opened 2018-09-25
compat: do not stop rpcbind service when enabling nis


pbrezina:
This is a problem on a nis server. Stopping `rpcbind` will also stop `ypserv`. We do not need to restart this service, only make sure it is running.

PR #101, opened 2018-09-25
compat: do not stop rpcbind, only start it

pbrezina:
Stopping rpcbind on a nis server will also stop ypserv as it depends
on rpcbind. There is no need to actually restart this service so we
only make sure that it is up and running.

Resolves:
https://github.com/pbrezina/authselect/issues/100


pbrezina:
Pushed to master:
* ecebf425168a88e505cee77ead74c173180bc55b

PR #102, opened 2018-09-26
sssd: document that this profile can be used also with sssd disabled

pbrezina:
https://github.com/pbrezina/authselect/issues/99


pbrezina:
Pushed to master:
* 0f6e2d29828580b75977feb5e0bde6d00c4eba3f

PR #105, opened 2018-10-08
sssd: support local users authentication with smartcards and add with-require-smartcard feature

pbrezina:


pbrezina:
@sumit-bose  Thank you for your review.

>In 'profiles/sssd/README' it is said that '"with-smartcard" must be set as well.' Wouldn't it be easier if 'with-require-smartcard' would implicitly set 'with-smartcard as well? Btw, the README doe not mention this for 'with-smartcard-lock-on-removal'.

Now it is mentioned also in lock on removal. At this moment, this is a limitation of the templating language.

> Maybe 'with-smartcard-required' would be a better name for the feature because then all Smartcard related features have the 'with-smartcard' prefix?

I agree, changed.

> For the required Smartcard authentication the dconf settting 'enable-password-authentication=false' should be added for gdm as well.

Done.

> If 'with-require-smartcard' is used with 'with-fingerprint' the PAM side of fingerprint authentication is disabled by adding 'pam_deny'. But in the gdm dconf setting 'enable-fingerprint-authentication

We hit template language limitation here as well. I will extend the templating language to support some logical operators as well but it is out of scope for beta release as it will take some time.

I opened https://github.com/pbrezina/authselect/issues/106 to track it.

sumit-bose:
Hi @pbrezina , thanks for the update. Patches are looking good now. ACK

pbrezina:
Pushed to master:
* 088a2b92742cab5e1d8f71452c2ae0c0f183a6fd
* 396089c2acc76bef59040d22c4170673ac4009bf

Issue #112, opened 2018-11-12
compat: recommend oddjob-mkhomedir


pbrezina:
Original authconfig used to choose oddjob-mkhomedir only if it is installed and used simple pam_mkhomedir otherwise. Authselect uses only oddjob version of mkhomedir. Therefore we should make sure that it is installed on the system when --enablemkhomedir is passed to the compat tool.

We can install the package conditionally, but this would mean that we have to require python-dnf which is actually much bigger dependency then simply requiring oddjob-mkhomedir. Given that this module is quite common, I incline to the simple requirement in spec file.

PR #113, opened 2018-11-12
spec: recommend oddjob-mkhomedir for compat tool

pbrezina:
Original authconfig used to choose oddjob-mkhomedir only if it is installed and used simple pam_mkhomedir otherwise. Authselect uses only oddjob version of mkhomedir. Therefore we should make sure that it is installed on the system when --enablemkhomedir is passed to the compat tool.

We can install the package conditionally, but this would mean that we have to require python-dnf which is actually much bigger dependency then simply requiring oddjob-mkhomedir. Given that this module is quite common, I incline to the simple requirement in spec file.

Resolves:
https://github.com/pbrezina/authselect/issues/112

### DIFF
--- a/rpm/authselect.spec.in
+++ b/rpm/authselect.spec.in
@@ -59,10 +59,10 @@ Obsoletes: authconfig < 7.0.1-6
 Provides: authconfig
 BuildRequires: python3-devel
 Requires: authselect%{?_isa} = %{version}-%{release}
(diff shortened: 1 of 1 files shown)


pbrezina:
Pushed to master:
* c4912a1f26a700f507ee5d21711ce1659278e736

Pushed to 1.0.x:
* 31117c1faec6d6d653f0b1c1e5959c23d86ed830

PR #124, opened 2018-11-30
profile: add 'default' profile RHBZ#1654018

jcpunk:
This adds a basic profile which uses only local users and only default pam modules.  There is an option to add additional features using non-default modules.

This can provide a simple template for folks looking to generate their own profiles as well.

### DIFF
--- a/profiles/Makefile.am
+++ b/profiles/Makefile.am
@@ -1,6 +1,19 @@
 authselect_profile_dir=@AUTHSELECT_PROFILE_DIR@
 
 # The shipped profiles
+profile_defaultdir = $(authselect_profile_dir)/default
+dist_profile_default_DATA = \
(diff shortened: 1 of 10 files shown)


pbrezina:
Thank you for your pull request and I'm sorry for such late answer. Most Fedora users do not have need for such profile (except those very few and experienced that want to use Fedora in non-default configuration without SSSD). Because each profile adds additional maintenance burden, we want to keep the number of upstream profiles as low as possible. For this reason, I must reject this pull request for now. If the number of users asking for this profile will increase in time, we will reconsider including it. 

Thank you for understanding.

jcpunk:
I am sad to hear this.  I was hopeful this profile would be included and could be used as a basis for downstream customizations.

PR #127, opened 2018-12-18
fscrypt support

antonblanchard:
I took a look at adding fscrypt support to Fedora (for ext4 encryption). These patches are enough to get it going, apart from some SELinux issues I need to look at next.

I'd appreciate feedback on the keyinit change, I wonder if I should have added it to the systemd-user file directly instead of system-auth which is included by a lot of stuff.

### DIFF
--- a/profiles/nis/README
+++ b/profiles/nis/README
@@ -47,6 +47,9 @@ with-nispwquality::
 without-nullok::
     Do not add nullok parameter to pam_unix.
 
+with-fscrypt::
+    Enable automatic per-user fscrypt based encryption.
+
 DISABLE SPECIFIC NSSWITCH DATABASES
 -----------------------------------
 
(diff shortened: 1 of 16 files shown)


pbrezina:
Thank you for your pull request, I'll add it on my todo list.

antonblanchard:
Thanks @pbrezina! FYI there were other things I had to do:

- Add /etc/pam.d/fscrypt, otherwise when fscrypt calls into pam it hits the default deny handler.
- Set SELinux to permissive mode, I ran out of time to look at exactly what fscrypt needs here.

There's discussion over in https://github.com/google/fscrypt/issues/111 on documenting fscrypt setup on various distros, perhaps they can help.

pbrezina:
Is fscrypt available in Fedora? I can not find this package.

pbrezina:
`fscrypt` is not available neither in recently released Fedora 30. I would like to include this module in upstream, but it does not make much sense because authselect is used only in Fedora and RHEL at the moment and both distros do not ship this module. I'm closing this pull request for now, please reopen it once the situation changes.

Issue #121, opened 2018-11-27
do not require root priviledges where it is not required


pbrezina:
Some authselect commands do not need root priviledges, thus we should not require it.

Bugzilla:
https://bugzilla.redhat.com/show_bug.cgi?id=1653718

pbrezina:
master:
* 1b407a9efe846fac69e83af0005b8a811bfa4eb9

PR #129, opened 2019-01-07
authselect: require root only for commands where it makes sense

pbrezina:
With this patch, only the following commands require root priviledges
because they require write access to system locations:

- select
- apply-changes
- enable-feature
- disable-feature
- create-profile

Resolves:
https://github.com/pbrezina/authselect/issues/121

### DIFF
--- a/src/cli/cli_tool.c
+++ b/src/cli/cli_tool.c
@@ -20,6 +20,7 @@
 
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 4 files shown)


thalman:
Could you resolve the conflict?

pbrezina:
rebased

pbrezina:
master:
* 1b407a9efe846fac69e83af0005b8a811bfa4eb9

Issue #128, opened 2018-12-21
Use setfscreatecon to create content in /etc/authselect with a proper context


doverride:

doverride:
https://www.systutorials.com/docs/linux/man/3-setfscreatecon/

Because the alternative would be to run restorecon -RvF /etc/authselect (or equivalent) to ensure proper labeling and that would be racy.

files like /etc/authselect/nsswitch.conf need to be labeled properly, and you create them with a random suffix so there is no other way to ensure proper labels.

doverride:
I seem to recall that at some point you actually did this properly but that is no longer the case, and i cannot find any trace of any changes in this regard in your git log

doverride:
it was probably authconfig-compat that did this properly...

pbrezina:
Thanks, I will look into it.

pbrezina:
PR: https://github.com/pbrezina/authselect/pull/131

PR #131, opened 2019-01-10
lib: label temporary files with correct selinux context

pbrezina:
Resolves:
https://github.com/pbrezina/authselect/issues/128

### DIFF
--- a/src/lib/util/selinux.c
+++ b/src/lib/util/selinux.c
@@ -1,0 +1,143 @@
+/*
+    Authors:
+        Pavel Březina <pbrezina@redhat.com>
+
+    Copyright (C) 2018 Red Hat
+
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 3 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#include <errno.h>
+#include <string.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <selinux/selinux.h>
+#include <selinux/label.h>
+
+#include "common/common.h"
+
+errno_t
+selinux_get_default_context(const char *path,
+                            char **_context)
+{
+    struct selabel_handle *handle;
+    char *context;
+    errno_t ret;
+
+    handle = selabel_open(SELABEL_CTX_FILE, NULL, 0);
+    if (handle == NULL) {
+        ret = errno;
+        ERROR("Unable to create selable context [%d]: %s", ret, strerror(ret));
+        return ret;
+    }
+
+    ret = selabel_lookup(handle, &context, path, 0);
+    if (ret != 0) {
+        ret = errno;
+        if (ret == ENOENT) {
+            return ENOENT;
+        }
+
+        ERROR("Unable to lookup selinux context [%d]: %s", ret, strerror(ret));
+    } else {
+        *_context = context;
+        ret = EOK;
+    }
+
+    selabel_close(handle);
+
+    return ret;
+}
+
+errno_t
+selinux_mkstemp_of(const char *filepath,
+                   char **_tmpfile)
+{
+    char *original_context = NULL;
+    char *default_context = NULL;
+    char *tmpfile = NULL;
+    errno_t ret;
+    int seret;
+    int fd;
+
+    seret = getfscreatecon(&original_context);
+    if (seret != 0) {
+        ERROR("Unable to get current fscreate selinux context!");
+        return EIO;
+    }
+
+    tmpfile = format("%s.XXXXXX", filepath);
+    if (tmpfile == NULL) {
+        ret = ENOMEM;
+        goto done;
+    }
+
+    ret = selinux_get_default_context(filepath, &default_context);
+    if (ret == ENOENT) {
+        default_context = NULL;
+    } else if (ret != EOK) {
+        ERROR("Unable to get default selinux context for [%s] [%d]: %s!",
+              filepath, ret, strerror(ret));
+        goto done;
+    }
+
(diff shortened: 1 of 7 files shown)


pbrezina:
Authselect first generates temporary files, then rename them to their proper name under /etc/authselect. This unfortunately creates them with a wrong selinux context. With this patch, correct context is used. You can test it with `ls -lZ /etc/authselect`.

src/lib/util/selinux.c
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    seret = setfscreatecon(default_context);
+    if (seret != 0) {
+        ERROR("Unable to set fscreate selinux context!");
+        ret = EIO;
+        goto done;
+    }
+
+    fd = mkstemp(tmpfile);
```
jhrozek: Normally calls don't set errno to 0 on success, don't you want to set ret = errno only on failure?

src/lib/util/selinux.c
```diff
@@ -0,0 +1,136 @@
...
+    ret = errno;
+
+    seret = setfscreatecon(original_context);
```
jhrozek: Why do you call this even if mkstemp fails?

src/lib/util/selinux.c
```diff
@@ -0,0 +1,136 @@
...
+    if (seret != 0) {
+        ERROR("Unable to restore fscreate selinux context!");
+        ret = EIO;
```
jhrozek: Don't you leak fd here?

src/lib/util/selinux.c
pbrezina: Right. I tried to find a nice way to call `setfscreatecon` only once to restore the original context, but perhaps there is not any.

src/lib/util/selinux.c
pbrezina: To restore original default selinux context for new files.

src/lib/util/selinux.c
pbrezina: Yes, I do.

pbrezina:
Thank you. I pushed fixed patch.

doverride:
Hello and thanks for addressing it. Even though I cannot C, I just wanted to just give some more info just to clarify:

For the sake of example, here is a test to see if the result is as expected (i am using type passwd_file_t for the sake of example:

# Tell selinux that /etc/authselect/nsswitch.conf should have type passwd_file_t:
echo "(filecon "/etc/authselect/nsswitch\.conf" file (system_u object_r passwd_file_t ((s0)(s0))))" > mytest.cil

# load it into the policy db:
semodule -i mytest.cil

# verify that selinux got the message:
matchpathcon /etc/authselect/nsswitch.conf

# rm the existing /etc/authselect and make authselect recreate it
# verify that authselect created it with the type we specified:

ls -alZ /etc/authselect/nsswitch.conf
now running restorecon should not try to change it because its right:
restorecon /etc/authselect/nsswitch.conf

# end of test remove mytest.cil: semodule -r mytest && rm mytest.cil

doverride:
If that does not do it for you, you can mke it  a bit broader (handy also for backups)

echo '(filecon "/etc/authselect/nsswitch\\.conf\\..*" file (system_u object_r passwd_file_t ((s0)(s0))))' >> mytest.cil

jhrozek:
Code-wise LGTM, but I haven't tested the code at all. Maybe it would be nice of @doverride could test some test build?

doverride:
Yes if you point me to a link to a scratch built on koji then i can test it but the example above should be easy enough to use to test it yourself

pbrezina:
@doverride Are you sure it should be `system_u object_r passwd_file_t`? If I call `sudo restorecon -RvF /etc/nsswitch.conf` it is `system_u:object_r:etc_t:s0`.

doverride:
Its just an example to test whether it works as expected

doverride:
if you add the filecon:

```
echo '(filecon "/etc/authselect/nsswitch\.conf\..*" file (system_u object_r passwd_file_t ((s0)(s0))))' >> mytest.cil
semodule -i mytest.cil
```
then restorecon will try to reset it to passwd_file_t

pbrezina:
Right, I understand it now. Thank you.

pbrezina:
I pushed a new version. The original version only relabeled the generated files under /etc/authselect but the symlinks still got wrong context. Now even symlinks (/etc/nsswitch.conf, /etc/pam.d/system-auth...) should have a proper context.

doverride:
How could you tell that the symlinks where mislabeled? My example ignored the symlinks.
Also the symlinks arent a problem. The problem was that authselect created the files in /etc/authselect with random suffixes and then renamed the files to the proper names.

This does not apply to the symlinks. The symlinks arent randomly suffixed and so we can deal with those in the policy.

So i would ignore tthe symlinks and only focus on the files in /etc/authselect

doverride:
You see, we can do a lot of labeling with policy. Its just that we cannot deal with files with random names in many cases. If the name of the file (or symlink) can be anticipated then policy can (and should preferably) be used. You want as little selinux awareness in your code as possible, but enough to make it work.

pbrezina:
@doverride In my test, `/etc/nsswitch.conf` was a symlink to `/etc/authselect/nsswitch.conf`. Its label was unconfined_u:object_r:etc_t. After running restorecon, it was relabeled to `system_u:object_r:etc_t` so I assume the label was not correct.

Here is a scratch build that sets the label for generated files and symlinks:
https://koji.fedoraproject.org/koji/taskinfo?taskID=31939960

If it is not necessary to set the label on symlink even if it does not match the result from calling restorecon, I can remove the code.

The algorithm works as:
1. Use selabel_lookup to obtain the default label for a file without tmp suffix.
2. Use setfscreatecon to set this label on new files
3. Create temporary file
4. Restore original label for setfscreatecon

Same for symlinks.

doverride:
Yes please remove the code for the symlink. The issue you describe is not related. It is because you use restorecon with -F (which is special) I will test the scratch built and report back

pbrezina:
Ok. I removed the code for symlinks and here is a new scratch build: https://koji.fedoraproject.org/koji/taskinfo?taskID=31940117

doverride:
The first one works (disregarded the symlink aspect) I will look at the second built as well

pbrezina:
Thank you.

Pushed to master:
* 9451dd55c43ce704e520dd4d62676466cc68b71e

Pushed to 1.0.x:
* 9c91693833c11af3793a87265108aa8094261d1a

Issue #134, opened 2019-02-04
sssd: require smartcard prohibits also password authentication via su or sudo


pbrezina:
Not all services should be forced to use smartcards.

Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1665058

pbrezina:
master:
* ed7b972a4b2a0c2a5874728bcd65e0df8ba05fd5

1.0.x:
* 8eb6b4c47514194bc741567b13452e216d4c27e2

PR #135, opened 2019-02-04
sssd: require smartcard only for specific services

pbrezina:
Otherwise even services like su or sudo can not perform password authentication
which is not desired.

Resolves:
https://github.com/pbrezina/authselect/issues/134


sumit-bose:
ACK, now Smartcards are required to get into the system but when logged in su or sudo can use passwords.

pbrezina:
Thank you. Pushed to master and 1.0.x.

Issue #136, opened 2019-02-14
selinux: if destination file exist use its label


pbrezina:
Currently we use default selinux label when we creating temporary file which is later renamed to its destination. However, if the destination already exist, we should use the current label.

pbrezina:
* `master`:
  * 6d6a1c666375a08f14c043bfe728df99f625a826
* `1.0.x`:
  * 3ac99e69058fabae19a3ddb7cc5f545350ad4046

PR #139, opened 2019-02-21
lib: use current selinux context if file already exist

pbrezina:
If the destination file exist we will use its selinux context for the
newly generated file. If it does not exist we will use its default
context.

You can set selinux context e.g. with `chcon`, e.g.:
  chcon -t httpd_sys_content_t /etc/authselect/nsswith.conf

And then check that is is kept with authselect apply-changes.

Resolves:
https://github.com/pbrezina/authselect/issues/136

### DIFF
--- a/src/lib/util/selinux.c
+++ b/src/lib/util/selinux.c
@@ -34,7 +34,7 @@ selinux_get_default_context(const char *path,
 {
     struct selabel_handle *handle;
     char *context;
-    errno_t ret;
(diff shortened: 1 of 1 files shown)


pbrezina:
* `master`:
  * 6d6a1c666375a08f14c043bfe728df99f625a826
* `1.0.x`:
  * 3ac99e69058fabae19a3ddb7cc5f545350ad4046

Issue #146, opened 2019-04-11
race condition in creating copy of current configuration


pbrezina:
We write config file to `/etc/authselect` and their copy to `/var/lib/authselect` so we can detect changes in the configuration. However, we do not create copy of the files in `/etc/authselect` but rather recreate them (see `authselect_system_write`). This has the side effect that the authselect preambule is regenerated and it may contain different time.

RH BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1697267

pbrezina:
Steps to reproduce:

```bash
$ sudo gdb -quiet authselect -ex 'set breakpoint pending on' -ex 'b src/lib/files/system.c:428' -ex 'run select sssd --force' -ex 'shell sleep 1' -ex 'detach' -ex 'quit'
$ sudo authselect check
[error] [/dev/shm/authselect-install/etc/authselect/system-auth] has unexpected content!
Current configuration is not valid. It was probably modified outside authselect.
```

pbrezina:
* `master`
  * 76a8cc9449cf8c1caeffc71a75c0760236ca2bba - lib: ensure that preambule timestamp is the same for all generated files
* `1.0.x`
  * c23f62c934ab2a620d0af60603e2fc3d1999ccd3 - lib: ensure that preambule timestamp is the same for all generated files

PR #151, opened 2019-04-25
lib: ensure that preambule timestamp is the same for all generated files

pbrezina:
To avoid possible race condition for configuration (/etc/authselect)
and state files (/var/lib/authselect).

Resolves:
https://github.com/pbrezina/authselect/issues/146

Steps to reproduce:
```
$ sudo gdb -quiet authselect -ex 'set breakpoint pending on' -ex 'b src/lib/files/system.c:428' -ex 'run select sssd --force' -ex 'shell sleep 1' -ex 'detach' -ex 'quit'
$ sudo authselect check
[error] [/dev/shm/authselect-install/etc/authselect/system-auth] has unexpected content!
Current configuration is not valid. It was probably modified outside authselect.
```

### DIFF
--- a/src/lib/files/system.c
+++ b/src/lib/files/system.c
@@ -361,12 +361,13 @@ done:
 static errno_t
 authselect_system_write_temp(const char *path,
                              const char *content,
+                             time_t timestamp,
                              char **_tmp_file)
 {
     errno_t ret;
 
     INFO("Writing temporary file for [%s]", path);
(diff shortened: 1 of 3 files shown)


pbrezina:
* `master`
  * 76a8cc9449cf8c1caeffc71a75c0760236ca2bba - lib: ensure that preambule timestamp is the same for all generated files
* `1.0.x`
  * c23f62c934ab2a620d0af60603e2fc3d1999ccd3 - lib: ensure that preambule timestamp is the same for all generated files

Issue #153, opened 2019-04-30
add list-features command


pbrezina:
Add a command to list all available features in given profile.

```
$ authselect list-features sssd
with-smartcard
with-sudo
with-mkhomedir
...
```

pbrezina:
* `master`
  * 215608a11df8857059e290e13d8fc53b35074150 - cli: add list-features command

PR #154, opened 2019-04-30
cli: add list-features command

pbrezina:
This command will print all features that are available within a profile.
It parses the profile templates and looks for features there so it does
not depend on the readme file.

Resolves:
https://github.com/pbrezina/authselect/issues/153

### DIFF
--- a/src/cli/main.c
+++ b/src/cli/main.c
@@ -400,6 +400,46 @@ done:
     return ret;
 }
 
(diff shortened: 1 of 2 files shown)


pbrezina:
* `master`
  * 215608a11df8857059e290e13d8fc53b35074150 - cli: add list-features command

Issue #106, opened 2018-10-09
Extend profile template language to support logical operators


pbrezina:
We hit language limitations in https://github.com/pbrezina/authselect/pull/105 and we should address them.

pbrezina:
* `master`
  * 1004ca36592473b859951586bc4a484040661a54 - lib: allow logical operators in templates

PR #155, opened 2019-05-06
tests: add template tests

thalman:
There are no unittests for template parser. It would be good
to have them before we extend the template language so we
can be sure that we will not break current behavior.

Part of #106

### DIFF
--- a/src/tests/Makefile.am
+++ b/src/tests/Makefile.am
@@ -16,6 +16,7 @@ noinst_HEADERS = \
 
 TESTS = \
     test_util_string_array \
+    test_util_template \
     $(NULL)
 
 check_PROGRAMS = $(TESTS)
@@ -29,5 +30,22 @@ test_util_string_array_CFLAGS = \
     $(AM_CFLAGS)
 test_util_string_array_LDADD = \
     $(CMOCKA_LIBS) \
+    $(top_builddir)/src/common/libcommon.la \
+    $(NULL)
+
+test_util_template_SOURCES = \
+    test_util_template.c \
+    ../lib/util/file.c \
+    ../lib/util/selinux.c \
+    ../lib/util/string.c \
+    ../lib/util/string_array.c \
+    ../lib/util/template.c \
+    ../lib/util/textfile.c \
+    $(NULL)
+test_util_template_CFLAGS = \
+    $(AM_CFLAGS)
+test_util_template_LDADD = \
+    $(CMOCKA_LIBS) \
+    $(SELINUX_LIBS) \
     $(top_builddir)/src/common/libcommon.la \
     $(NULL)
--- a/src/tests/test_util_template.c
+++ b/src/tests/test_util_template.c
@@ -1,0 +1,181 @@
+/*
+    Authors:
+        Tomas Halman <thalman@redhat.com>
+
+    Copyright (C) 2019 Red Hat
+
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 3 of the License, or
+    (at your option) any later version.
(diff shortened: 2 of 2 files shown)


pbrezina:
Thank you.

By avoiding unused variable you actually triggered another warning:
```
/home/pbrezina/workspace/authselect/src/tests/test_util_template.c: In function ‘textfile_write’:
/home/pbrezina/workspace/authselect/src/tests/test_util_template.c:31:5: error: statement with no effect [-Werror=unused-value]
     filepath; /* avoid not-used */
     ^~~~~~~~
/home/pbrezina/workspace/authselect/src/tests/test_util_template.c:32:5: error: statement with no effect [-Werror=unused-value]
     content;
     ^~~~~~~
/home/pbrezina/workspace/authselect/src/tests/test_util_template.c:33:5: error: statement with no effect [-Werror=unused-value]
     mode;
     ^~~~
/home/pbrezina/workspace/authselect/src/tests/test_util_template.c: In function ‘selinux_mkstemp_for’:
/home/pbrezina/workspace/authselect/src/tests/test_util_template.c:43:5: error: statement with no effect [-Werror=unused-value]
     filepath; /* avoid not-used */
     ^~~~~~~~
/home/pbrezina/workspace/authselect/src/tests/test_util_template.c:44:5: error: statement with no effect [-Werror=unused-value]
     mode;
     ^~~~
/home/pbrezina/workspace/authselect/src/tests/test_util_template.c:45:5: error: statement with no effect [-Werror=unused-value]
     _tmpfile;
```

This can be simply avoided by building the test properly with all dependencies. Which is more then fine, because I would like to also test other functions including `template_write` which brings these dependencies in.  (not necessarily in this PR)
```
[pbrezina ~/workspace/authselect (review *)]$ git diff
diff --git a/src/tests/Makefile.am b/src/tests/Makefile.am
index 132db08..0bfe356 100644
--- a/src/tests/Makefile.am
+++ b/src/tests/Makefile.am
@@ -35,13 +35,17 @@ test_util_string_array_LDADD = \
 
 test_util_template_SOURCES = \
     test_util_template.c \
-    ../lib/util/template.c \
-    ../lib/util/string_array.c \
+    ../lib/util/file.c \
+    ../lib/util/selinux.c \
     ../lib/util/string.c \
+    ../lib/util/string_array.c \
+    ../lib/util/template.c \
+    ../lib/util/textfile.c \
     $(NULL)
 test_util_template_CFLAGS = \
     $(AM_CFLAGS)
 test_util_template_LDADD = \
     $(CMOCKA_LIBS) \
+    $(SELINUX_LIBS) \
     $(top_builddir)/src/common/libcommon.la \
     $(NULL)
```

Please, avoid using global variables. If needed, cmocka provides setup and teardown functions that can create/destroy the state which can be then used inside tests. However, it is not required in this cases. It would be much better to define features in each test so it is clear what values are used for testing.

Also please, test only one thing in each test. I.e. split the tests into "if", "if not", "exclude if", "include if", ... instead of merging them to single unit tests. This way it will be more clear what the test does and what fails.

thalman:
Thanks for review, PR updated

pbrezina:
Ack. Thank you.

pbrezina:
I slightly changed the commit message (added a full URL to the ticket) and renamed the test to include complete keyword (i.e. `exclude_if` instead of just `exclude`). I went ahead and pushed it like this, hoping you do not mind.

* `master`
  * a089075279192c07ecfacec12992629a31f816fc - tests: add template tests

pbrezina:
* `master`
  * a089075279192c07ecfacec12992629a31f816fc - tests: add template test

thalman:
Good thanks. I will update the PR#158 of more template tests.

PR #158, opened 2019-05-09
lib: Allow logical operators in templates

thalman:
Templates are not flexible enough when we can use only one feature
in conditions. In some cases we need to logically combine two or
more features.

Here we implement logical operators (or, and, not).

Conditional line operator "if not" has been removed because "if"
operator is now sufficient.

Resolves: #106

### DIFF
--- a/src/lib/Makefile.am
+++ b/src/lib/Makefile.am
@@ -20,6 +20,7 @@ noinst_HEADERS = \
     util/string_array.h \
     util/string.h \
     util/template.h \
+    util/evaluator.h \
     util/textfile.h \
     util/util.h \
     $(NULL)
@@ -63,6 +64,7 @@ libauthselect_la_SOURCES = \
     util/string_array.c \
     util/string.c \
     util/template.c \
+    util/evaluator.c \
     util/textfile.c \
     $(NULL)
 libauthselect_la_LIBADD = \
--- a/src/lib/util/evaluator.c
+++ b/src/lib/util/evaluator.c
@@ -67,6 +67,11 @@ static struct e_map operators[] = {
     {NULL, E_OPERATOR_INVALID}
 };
 
+static errno_t evaluator_state_machine(struct evaluator *self,
+                                       int depth,
+                                       const char **features,
+                                       bool *_result);
+
 /*
  * This function reads new item/token from expression.
  * It is then stored in self->token. Expected tokens are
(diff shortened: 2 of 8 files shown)


pbrezina:
* only `evaluate();` is used outside `evaluator.c`, other function should be marked `static` and removed from `evaluator.h`
* `struct evaluator_t` should be accessible only from `evaluator.c`, please remove it from the header file
* please rename `struct evaluator_t` to `struct evaluator` (without `_t` suffix). The suffix is used by posix types  to avoid potential collisions with other tokens. However it does not make any sense to be used in structure definition (since keyword `struct` already differentiates it from other tokens) nor with project specific types. Same goes for other types that you define in the source file.
* please, use `malloc_zero()` from `common/common.h` instead of `malloc()` and `memset()`
* please, use `if (pointer == NULL)` instead of `if (!pointer)` and `if (pointer != NULL)` instead of `if (pointer)`
* please, refer to `features` instead of `variables`
* it is not semantically correct to refer to `struct evaluator_t` as `self` in `evaluate()`, your are not yet inside evaluator "object", you are creating it here
* `static int evaluator_token_to_state(const char *token) {` please, keep opening bracket on the new line
* `static int evaluator_token_to_state()` (an possibly other functions) do not return int but proper enumeration type `enum ...`
* `evaluator_operator()` use `for` instead of `while`
* the evaluator should return result as boolean instead of integer
* `struct evaluator_t` field `token` does not have to be part of this structure, it is only set but never read
* `evaluator_next_token()` use `isblank()` instead of `*self->cursor == ' ' || *self->cursor == '\t'`
* `evaluator_next_token()`: return `errno_t` from the function and return the token as an output parameter to avoid using `'\1'` to indicate error
* `evaluator_get_variable()` return `errno_t` instead of `int` and `EOK` instead of `0`
* `if (ret != EOK) return ret;` please always write even simple if on multiple lines with brackets i.e. `if (..) {\nreturn ret;\n}`

thalman:
I addressed all comment except that one about `token` being member of `evaluator`. It is/was used but now after update it is more visible.

pbrezina:
Thank you, this is better.

* `bool evaluator_state_machine()` it returns `errno_t`, not `bool`
* The code breaks authselect completely because. It breaks `template_list_features()` and therefore even `authselect select` does not work. The function will return list of expressions instead of list of  features. I think it would be enough to use different regular expressions there, something that will parse `{(keyword) ... "feature" ... "feature" ... }` and return list of feature names.

Here are some nitpicks:
* There are several coding style issues. The coding style is not yet documented but we used the same style as in SSSD. Some differences would be OK if you kept the style consistent inside a module, but you keep changing it from line to line. So please:
    * do not add space after negation: `! sub_result` -> `!sub_result`
    * add space between operators: `strlen(expression)+1` -> `strlen(expression) + 1`
    * keep 80 characters per line limit
* `evaluator_free()` is over complicated, the code will be much simpler if you just inline it in `evaluate()` where it is used. You can also inline `evaluator_new()` which is a one liner anyway. This would make sense if it was a public API, but since those functions are only internal for the module it just adds unnecessary code.
* I would appreciated some comments in `evaluator_next_token()` and `evaluator_token_to_state()`. I do not mean commenting what the code do, but some examples of what tokens do you read/parse there would be nice.
* Would it be possible to split `evaluator_state_machine()` into more functions to improve readability?

pbrezina:
Also, given you are a Red Hatter you should add Red Hat copyright to the file, see other files.

thalman:
Here are incorporated comments.

I also tried to break state machine into more functions. Take a look, I'm not sure of it so I keep it as separate commit so far.

pbrezina:
Thank you. Ack.

* `master`
  * 1004ca36592473b859951586bc4a484040661a54 - lib: allow logical operators in templates

PR #160, opened 2019-05-10
Add krb5 and logon cache in winbind profile

mikhailnov:


profiles/winbind/README
```diff
@@ -48,6 +48,12 @@ with-pam-u2f-2fa::
 with-krb5::
     Enable Kerberos authentication with *pam_winbind*.
 
+with-winbind-cache::
+    Enable authentication cache with *pam_winbind*.
```
cryptomilk: I would write:

Enable logon using cached credentials with *pam_winbind* when 'winbind offline logon' is enabled.

profiles/winbind/password-auth
```diff
@@ -5,7 +5,7 @@ auth        sufficient                                   pam_u2f.so cue
 auth        required                                     pam_u2f.so cue nouserok                                  {include if "with-pam-u2f-2fa"}
 auth        sufficient                                   pam_unix.so {if not "without-nullok":nullok} try_first_pass
 auth        requisite                                    pam_succeed_if.so uid >= 1000 quiet_success
-auth        sufficient                                   pam_winbind.so {if "with-krb5":krb5_auth} use_first_pass
+auth        sufficient                                   pam_winbind.so {if "with-krb5":krb5_auth} {if "with-krb5-cache":krb5_ccache_type=KEYRING} {if "with-winbind-cache":cached_login} use_first_pass
```
cryptomilk: This needs to be solved differently as the new default cache is KCM and not KEYRING anymore! So you need to allow to set the type via the commandline option.

pbrezina:
Hi. I|m sorry for the delay, there was a lot going on lately. `cached_login` needs to be enabled also in `smb.conf`.  I think it would be nice to note this in the `REQUIREMENTS` file. I think thing about the KCM option mentioned by Andreas above.

pbrezina:
I see this in map pam_winbind:
```
       krb5_ccache_type=[type]
           When pam_winbind is configured to try kerberos authentication by enabling the krb5_auth option, it can store the retrieved Ticket Granting Ticket (TGT) in a credential cache. The type of credential cache can be set with
           this option. Currently the only supported value is: FILE. In that case a credential cache in the form of /tmp/krb5cc_UID will be created, where UID is replaced with the numeric user id. Leave empty to just do kerberos
           authentication without having a ticket cache after the logon has succeeded.
```
Specifically: *Currently the only supported value is: FILE*

@cryptomilk Is the manual page outdated?

cryptomilk:
Yes, it is outdated. FILE, WRFILE, DIR and KEYRING are currently supported. I have to add KCM asap.

Could you open a bug that it isn't documented?

pbrezina:
@cryptomilk Sure, will do. What is the default value? FILE?

cryptomilk:
The default is FILE, yes.

pbrezina:
@cryptomilk So the upstream profiles should follow the Fedora defaults and it can be changed to something else on other distributions. Given the default value will change to KCM wchi AFAIK is current default for Kerberos, I think it is not needed to add this option to upstream profile at all. Does it make any sense to switch form KCM to KEYRING (other then user's personal preference)?

@mikhailnov If I understand it correctly, ROSA is based on RHEL. Is it possible that you will adopt KCM as default as well?

pbrezina:
> Could you open a bug that it isn't documented?

For the record, I opened: https://bugzilla.redhat.com/show_bug.cgi?id=1714963

mikhailnov:
> @mikhailnov If I understand it correctly, ROSA is based on RHEL. Is it possible that you will adopt KCM as default as well?

Sorry for very late responce. Yes, there is an RHEL-based ROSA, but this ROSA is a fork of Mandriva, ROSA bought Mandriva company. It' s a completely independent distribution.

Thank you for all your comments, seems that documentation of samba has not been up[dated for the new default ccache type, could you please clearify which type is recommended to use with samba 4.10.x?

pbrezina:
> Thank you for all your comments, seems that documentation of samba has not been up[dated for the new default ccache type, could you please clearify which type is recommended to use with samba 4.10.x?

@cryptomilk Do you know what ccache is recommended? Also does it have to even be in pam configuration? Isn't it sufficient to have it in pam_winbind.conf?

mikhailnov:
> Also does it have to even be in pam configuration? Isn't it sufficient to have it in pam_winbind.conf?

Good question.

> Do you know what ccache is recommended?

@cryptomilk will smbspool_krb5_wrapper be able to find e.g. KEYRING kerberos ccache by `krb5_cc_get_full_name()` ? If no, then FILE should be left as default I think.
Kerberos docs say:

"KEYRING is Linux-specific, and uses the kernel keyring support to store credential data in unswappable kernel memory where only the current user should be able to access it."

smbspool_krb5_wrapper runs from root and tries to get krb5 ccache of specified user, I think it will not work with KEYRING ccache.

pbrezina:
I'm going to close this PR because it should be sufficient to set these options in `pam_winbind.conf`.

Perhaps ask on [samba-technical](https://lists.samba.org/mailman/listinfo/samba-technical) to get a qualified answer to your questions since I can't answer that.

PR #161, opened 2019-05-10
Support libnss-role in /etc/nsswitch.conf

mikhailnov:
Enaple support of libnss-role (https://github.com/Etersoft/libnss-role),
which allows to map domain group to local ones, e.g. users from domain
group 'Domain Admins' may be automatically added to the local group 'wheel'.

This change breaks beautiful formatting of `nsswitch.conf`, but I did not want to include unneded chnages into the diff, hoewever, I may fix formatting.

### DIFF
--- a/profiles/nis/nsswitch.conf
+++ b/profiles/nis/nsswitch.conf
@@ -1,7 +1,7 @@
 aliases:    files nis                   {exclude if "with-custom-aliases"}
 automount:  files nis                   {exclude if "with-custom-automount"}
 ethers:     files nis                   {exclude if "with-custom-ethers"}
-group:      files nis systemd           {exclude if "with-custom-group"}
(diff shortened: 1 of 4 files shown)


pbrezina:
Hi, thank you for your contribution. Is this library available in Fedora/RHEL or is it not yet packaged in these distros?

mikhailnov:
Hi, I was sure that you would ask this ;) Unfortunately it is not according to https://pkgs.org/

pbrezina:
I would like to include this change in upstream, however since authselect is at this moment still Fedora-driven project, I must reject it. I am very happy that it is being adopted by other distribution, but it would be asking for troubles if we would include a module that is not available on all distros that adopted authselect. It is better to add distro-specific module in your distribution only than to keep removing it in Fedora/other future distributions. Thank you for understanding.

PR #162, opened 2019-05-14
Make test build more strict about warnings

thalman:
We want to know about warnings in the code and fix them.
Therefore we set environment variable CFLAGS

CFLAGS="-Wall -Wextra -Werror -Wno-sign-compare -Wno-unused-parameter -ggdb3"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,13 @@ sudo: false
 services:
 - docker
 
+env:
+  global:
+  - CFLAGS="-Wall -Wextra -Werror -Wno-sign-compare -Wno-unused-parameter -ggdb3"
+
 before_install:
 - docker pull fedora
-- docker run -d --rm --name AUTHSELECT-FEDORA -v `pwd`:/buildroot fedora /bin/sh -c "tail -f /dev/null"
+- docker run -d --rm --name AUTHSELECT-FEDORA -e CFLAGS -v `pwd`:/buildroot fedora /bin/sh -c "tail -f /dev/null"
 - docker ps
 - docker exec AUTHSELECT-FEDORA dnf -y install gcc make gettext gettext-devel autoconf automake libtool m4 pkgconfig popt-devel asciidoc libcmocka-devel libselinux-devel file rpm-build python3-devel po4a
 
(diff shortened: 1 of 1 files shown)


pbrezina:
Apparently there is [this option](https://blog.travis-ci.com/2013-11-27-fast-finishing-builds) that can be used instead:

```yaml
matrix:
  fast_finish: true
```

The rpm build fails because your `CFLAGS` do not generate debug information. I'm not sure what is the default `CFLAGS` value in the docket machine but it would be better to keep it. If it will not help, then you can add `-ggdb3` to the `CFLAGS` value instead of unsetting it. Something like:

```
env:
  global:
  - AUHTSELECT_CFLAGS="-Wall -Wextra -Werror -Wno-sign-compare -Wno-unused-parameter"

before_install:
    ...
    - docker run -d --rm --name AUTHSELECT-FEDORA -e AUTHSELECT_CFLAGS -v `pwd`:/buildroot fedora /bin/sh -c "echo CFLAGS=\"$CFLAGS $AUTHSELECT_CFLAGS\" >> ~/.bashrc"
```

thalman:
> 
> ```yaml
> matrix:
>   fast_finish: true
> ```
> 
Not really. This options is for "matrix" build (several builds running in parallel with different settings).
I found several discussion on this topic and it seems that solution is to do it as I suggest.

> The rpm build fails because your `CFLAGS` do not generate debug information. I'm not sure what is the default `CFLAGS` value in the docker machine but it would be better to keep it.
There is no default for `CFLAGS`, it is ours setting in `.travis.yml`.

 I will try with -g. Thanks for this hint

thalman:
PR updated, please review

pbrezina:
Thank you for explanation.  Please, just squash `-ggdb3 to the first patch. I will push it after this change.

thalman:
Done.

pbrezina:
Before I push it - is there any reason why you removed `docker stop AUTHSELECT-FEDORA` or is this a mistake?

thalman:
There were two docker commands actually without any real effect. This `docker stop` and also `docker ps` in `before_install` section. Docker service is stopped at the end of build with Travis container itself.

So removing it is intentional.

pbrezina:
Thank you.

pbrezina:
* `master`
  * bd5fc9ef878f50ee3e8349515a7252826656187c - Fail the build on first error
  * 0122a1eaa9ec920642e203b0aeed62299867a180 - Make test build more strict about warnings

Issue #170, opened 2019-05-28
lib: use length comparison to check for unique value in string_array_add_value_safe()


pbrezina:
Because the `value` in `string_array_add_value_safe()` does not have to end with `\0` terminator, we might fail to check if value already exist if `strlen(value) < len`.

pbrezina:
* `master`
  * ee88d10cd44f7d8e91b2ac14ddd3463894e02a26 - lib: use length comparison to check for unique value in string_array_add_value_safe()
  * 4a8a7b447c44333e9669917fe599c24168b1f365 - lib: add string_array_has_value_safe()
* `1.0.x`
  * c075ccaee89950a4182f3dff6bf8f94e7eca87db - lib: use length comparison to check for unique value in string_array_add_value_safe()
  * 2d5281dfdc0709b434bcf6852d27f337bc376b37 - lib: add string_array_has_value_safe()

PR #171, opened 2019-05-28
lib: use length comparison to check for unique value in string_array_add_value_safe()

pbrezina:
Because the `value` in `string_array_add_value_safe()` does not have to end 
with \0 terminator, we might fail to check if value already exist if
`strlen(value) < len`.

Resolves:
https://github.com/pbrezina/authselect/issues/170

### DIFF
--- a/src/lib/util/string_array.c
+++ b/src/lib/util/string_array.c
@@ -89,7 +89,7 @@ string_array_count(char **array)
 }
 
 bool
-string_array_has_value(char **array, const char *value)
+string_array_has_value_safe(char **array, const char *value, size_t len)
 {
     int i;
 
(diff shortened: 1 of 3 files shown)


src/lib/util/string_array.c
```diff
@@ -98,14 +98,20 @@ string_array_has_value(char **array, const char *value)
     }
 
     for (i = 0; array[i] != NULL; i++) {
-        if (strcmp(value, array[i]) == 0) {
+        if (strncmp(value, array[i], len) == 0) {
```
thalman: I think that this will fail when there is longer value with the same beginning in the array. Imagine
* have in array "hello"
* add "hell", 4

pbrezina: Fixed. Good catch.

pbrezina:
I pushed fixed patch.

pbrezina:
* `master`
  * ee88d10cd44f7d8e91b2ac14ddd3463894e02a26 - lib: use length comparison to check for unique value in string_array_add_value_safe()
  * 4a8a7b447c44333e9669917fe599c24168b1f365 - lib: add string_array_has_value_safe()
* `1.0.x`
  * c075ccaee89950a4182f3dff6bf8f94e7eca87db - lib: use length comparison to check for unique value in string_array_add_value_safe()
  * 2d5281dfdc0709b434bcf6852d27f337bc376b37 - lib: add string_array_has_value_safe()

Issue #166, opened 2019-05-21
sssd: dconf does not lock enable-password-authentication


pbrezina:
We set `enable-password-authentication` in `dconf-db` so it should be also locked in `dconf-locks`.

pbrezina:
* `master`
  * 105b0f072b9635a378adfcca1b4a4bda61f31665 - sssd: lock /org/gnome/login-screen/enable-password-authentication
* `1.0.x`
  * df39a71cc4ea1ef8b3afb52975999064233881aa - sssd: lock /org/gnome/login-screen/enable-password-authentication

PR #172, opened 2019-05-28
sssd: lock /org/gnome/login-screen/enable-password-authentication

pbrezina:
We set this option in dconf-db so we should also lock it.

Resolve:
https://github.com/pbrezina/authselect/issues/166

### DIFF
--- a/profiles/sssd/dconf-locks
+++ b/profiles/sssd/dconf-locks
@@ -1,3 +1,4 @@
 /org/gnome/login-screen/enable-smartcard-authentication
 /org/gnome/login-screen/enable-fingerprint-authentication
+/org/gnome/login-screen/enable-password-authentication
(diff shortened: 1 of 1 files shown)


pbrezina:
After the patch:
```
$ git grep enable-password-authentication
profiles/sssd/dconf-db:enable-password-authentication={if "with-smartcard-required":false|true}
profiles/sssd/dconf-locks:/org/gnome/login-screen/enable-password-authentication
```

pbrezina:
* `master`
  * 105b0f072b9635a378adfcca1b4a4bda61f31665 - sssd: lock /org/gnome/login-screen/enable-password-authentication
* `1.0.x`
  * df39a71cc4ea1ef8b3afb52975999064233881aa - sssd: lock /org/gnome/login-screen/enable-password-authentication

PR #174, opened 2019-05-29
cli: The --backup opt can't be used

thalman:
We can't detect properly usage of --backup
option without backup name.

### DIFF
--- a/src/cli/main.c
+++ b/src/cli/main.c
@@ -137,7 +137,8 @@ static errno_t activate(struct cli_cmdline *cmdline)
     const char **features = NULL;
     const char *profile_id;
     char *requirements = NULL;
-    char *backup_name = NULL;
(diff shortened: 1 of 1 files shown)


pbrezina:
Thank you for looking into it. Unfortunately this is not doable, because it is not possible to differentiate backup name from a feature that should be enabled, e.g.:

```
$ sudo ./authselect select sssd --backup with-sudo --force
Backup stored at /dev/shm/authselect-install/var/lib/authselect/backups/with-sudo
```

The feature names does not have to necessarily follow the `with-` syntax that we use in upstream profiles so we can not really assume what is supposed to be a feature name. Similarly, we could list features of the profile and remove these features from the arguments so we can parse them with popt, but again we can not assume that an unknown feature is a backup name and not a typo.

PR #184, opened 2019-10-17
Update faillock and pwquality options in pam configs

redhatrises:
- For faillock, it is now recommended to configure settings in /etc/security/faillock.conf
- local_users_only has been moved in libpwquality to /etc/security/pwquality.conf

### DIFF
--- a/profiles/nis/fingerprint-auth
+++ b/profiles/nis/fingerprint-auth
@@ -1,8 +1,8 @@
 {continue if "with-fingerprint"}
 auth        required                                     pam_env.so
-auth        required                                     pam_faillock.so preauth silent deny=4 unlock_time=1200 {include if "with-faillock"}
+auth        required                                     pam_faillock.so preauth silent                                {include if "with-faillock"}
 auth        sufficient                                   pam_fprintd.so
-auth        required                                     pam_faillock.so authfail deny=4 unlock_time=1200       {include if "with-faillock"}
+auth        required                                     pam_faillock.so authfail                               {include if "with-faillock"}
 auth        required                                     pam_deny.so
 
 account     required                                     pam_access.so                                          {include if "with-pamaccess"}
--- a/profiles/nis/password-auth
+++ b/profiles/nis/password-auth
@@ -1,11 +1,11 @@
 auth        required                                     pam_env.so
 auth        required                                     pam_faildelay.so delay=2000000
-auth        required                                     pam_faillock.so preauth silent deny=4 unlock_time=1200   {include if "with-faillock"}
+auth        required                                     pam_faillock.so preauth silent                                  {include if "with-faillock"}
 auth        sufficient                                   pam_u2f.so cue                                           {include if "with-pam-u2f"}
 auth        required                                     pam_u2f.so cue nouserok                                  {include if "with-pam-u2f-2fa"}
 auth        sufficient                                   pam_unix.so {if not "without-nullok":nullok} try_first_pass
 auth        requisite                                    pam_succeed_if.so uid >= 1000 quiet_success
-auth        required                                     pam_faillock.so authfail deny=4 unlock_time=1200         {include if "with-faillock"}
+auth        required                                     pam_faillock.so authfail                                 {include if "with-faillock"}
 auth        required                                     pam_deny.so
 
 account     required                                     pam_access.so                                            {include if "with-pamaccess"}
--- a/profiles/nis/system-auth
+++ b/profiles/nis/system-auth
@@ -1,12 +1,12 @@
 auth        required                                     pam_env.so
 auth        required                                     pam_faildelay.so delay=2000000
-auth        required                                     pam_faillock.so preauth silent deny=4 unlock_time=1200 {include if "with-faillock"}
+auth        required                                     pam_faillock.so preauth silent                                {include if "with-faillock"}
 auth        sufficient                                   pam_fprintd.so                                         {include if "with-fingerprint"}
 auth        sufficient                                   pam_u2f.so cue                                         {include if "with-pam-u2f"}
 auth        required                                     pam_u2f.so cue nouserok                                {include if "with-pam-u2f-2fa"}
 auth        sufficient                                   pam_unix.so {if not "without-nullok":nullok} try_first_pass
 auth        requisite                                    pam_succeed_if.so uid >= 1000 quiet_success
-auth        required                                     pam_faillock.so authfail deny=4 unlock_time=1200       {include if "with-faillock"}
+auth        required                                     pam_faillock.so authfail                               {include if "with-faillock"}
 auth        required                                     pam_deny.so
 
 account     required                                     pam_access.so                                          {include if "with-pamaccess"}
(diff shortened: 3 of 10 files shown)


redhatrises:
Should the authconfig-compat python code be updated as well?

pbrezina:
Thank you. I think there is nothing to change in the compat tool.

`local_users_only` should be kept for `sssd` and `winbind` profiles as password policies are managed remotely.

What are the default values when using faillock.conf. Is it the same as what we have in profiles?

redhatrises:
> Thank you. I think there is nothing to change in the compat tool.
> 
> `local_users_only` should be kept for `sssd` and `winbind` profiles as password policies are managed remotely.

@pbrezina `local_users_only` should now configured in `/etc/security/pwquality.conf` as of libpwquality-1.4.2. Is there code in authselect that can do this?

> What are the default values when using faillock.conf. Is it the same as what we have in profiles?

The defaults are `deny=3` and `unlock_time=900` which is actually quite standard these days. If they need to be different, is there code that can handle changing the configs in `faillock.conf`?

pbrezina:
> > Thank you. I think there is nothing to change in the compat tool.
> > `local_users_only` should be kept for `sssd` and `winbind` profiles as password policies are managed remotely.
> 
> @pbrezina `local_users_only` should now configured in `/etc/security/pwquality.conf` as of libpwquality-1.4.2. Is there code in authselect that can do this?

No, authselect does not touch other configuration files. I will think about the correct place for local_users_only and ask few people, but I still feel that it should be kept in the pam configuration at this moment.

> 
> > What are the default values when using faillock.conf. Is it the same as what we have in profiles?
> 
> The defaults are `deny=3` and `unlock_time=900` which is actually quite standard these days. If they need to be different, is there code that can handle changing the configs in `faillock.conf`? See https://www.pagure.io/pam-redhat/blob/master/f/pam_faillock/faillock.conf

I guess this is similar enough to what we have, but this change needs to be properly documented.

redhatrises:
> > > Thank you. I think there is nothing to change in the compat tool.
> > > `local_users_only` should be kept for `sssd` and `winbind` profiles as password policies are managed remotely.
> > 
> > 
> > @pbrezina `local_users_only` should now configured in `/etc/security/pwquality.conf` as of libpwquality-1.4.2. Is there code in authselect that can do this?
> 
> No, authselect does not touch other configuration files. I will think about the correct place for local_users_only and ask few people, but I still feel that it should be kept in the pam configuration at this moment.

@pbrezina just so you know, we are going to start flagging this in `scap-security-guide` so having this configured in `/etc/security/pwquality.conf` would be very helpful 

> 
> > > What are the default values when using faillock.conf. Is it the same as what we have in profiles?
> > 
> > 
> > The defaults are `deny=3` and `unlock_time=900` which is actually quite standard these days. If they need to be different, is there code that can handle changing the configs in `faillock.conf`? See https://www.pagure.io/pam-redhat/blob/master/f/pam_faillock/faillock.conf
> 
> I guess this is similar enough to what we have, but this change needs to be properly documented.

@pbrezina Where should the change be documented? `man pam_faillock` talks about the default setting.

pbrezina:
> > No, authselect does not touch other configuration files. I will think about the correct place for local_users_only and ask few people, but I still feel that it should be kept in the pam configuration at this moment.
> 
> @pbrezina just so you know, we are going to start flagging this in `scap-security-guide` so having this configured in `/etc/security/pwquality.conf` would be very helpful

I'm sorry, I did not got to this yet and I'm going on PTO tomorrow. I'll get back to it later.

> > > > What are the default values when using faillock.conf. Is it the same as what we have in profiles?
> > > 
> > > 
> > > The defaults are `deny=3` and `unlock_time=900` which is actually quite standard these days. If they need to be different, is there code that can handle changing the configs in `faillock.conf`? See https://www.pagure.io/pam-redhat/blob/master/f/pam_faillock/faillock.conf
> > 
> > 
> > I guess this is similar enough to what we have, but this change needs to be properly documented.
> 
> @pbrezina Where should the change be documented? `man pam_faillock` talks about the default setting.

Release notes is enough from upstream point of view. But since it is changed of default setting of a security related module, I think it will need a self-contained Fedora change page for downstream change.

redhatrises:
> > > No, authselect does not touch other configuration files. I will think about the correct place for local_users_only and ask few people, but I still feel that it should be kept in the pam configuration at this moment.
> > 
> > 
> > @pbrezina just so you know, we are going to start flagging this in `scap-security-guide` so having this configured in `/etc/security/pwquality.conf` would be very helpful
> 
> I'm sorry, I did not got to this yet and I'm going on PTO tomorrow. I'll get back to it later.
> 
> > > > > What are the default values when using faillock.conf. Is it the same as what we have in profiles?
> > > > 
> > > > 
> > > > The defaults are `deny=3` and `unlock_time=900` which is actually quite standard these days. If they need to be different, is there code that can handle changing the configs in `faillock.conf`? See https://www.pagure.io/pam-redhat/blob/master/f/pam_faillock/faillock.conf
> > > 
> > > 
> > > I guess this is similar enough to what we have, but this change needs to be properly documented.
> > 
> > 
> > @pbrezina Where should the change be documented? `man pam_faillock` talks about the default setting.
> 
> Release notes is enough from upstream point of view. But since it is changed of default setting of a security related module, I think it will need a self-contained Fedora change page for downstream change.

@pbrezina are you back yet from PTO to look at this? The latest releases of pam and libpwquality contain these changes and are now in Fedora 31.

t8m:
The local_users_only support in pwquality.conf is not yet provided and planned for RHEL-8.

pbrezina:
Thank you. I also talked with Tomas about keeping local_users_only in place to avoid running pwquality for LDAP users and he confirmed that we should keep it.

So please, keep `local_users_only` intact. And also keep `preauth silent`. Thank you.

redhatrises:
> So please, keep `local_users_only` intact. And also keep `preauth silent`. Thank you.

@pbrezina will keep `local_users_only` intact for now since it hasn't made its way into RHEL-8 yet.

`preauth silent` has been moved to `faillock.conf` as well hence its removal from here as well.
Updated the PR with the latest changes.

t8m:
> `preauth silent` has been moved to `faillock.conf` as well hence its removal from here as well.
> Updated the PR with the latest changes.

NO! `preauth` is NOT moved to faillock.conf. It cannot be.  The `silent` option can be set from faillock.conf but it is not the best idea to depend on faillock.conf in authselect for this particular thing as authselect does not modify faillock.conf, it is off by default, and `silent` option is necessary to not leak info about user being existing on the system or not.

redhatrises:
> > `preauth silent` has been moved to `faillock.conf` as well hence its removal from here as well.
> > Updated the PR with the latest changes.
> 
> NO! `preauth` is NOT moved to faillock.conf. It cannot be. The `silent` option can be set from 

@t8m apologies. This was a typo. `preauth` is wasn't moved from faillock.conf.

> faillock.conf but it is not the best idea to depend on faillock.conf in authselect for this particular thing as authselect does not modify faillock.conf, it is off by default, and `silent` option is necessary to not leak info about user being existing on the system or not.

Okay.... update the PR.

pbrezina:
Thank you. I fixed indentation and change the commit message title and pushed to master.

* `master`
  * efeb22eaf2a61b242b4318ae5569140855e3c508 - profiles: remove hardcoded arguments from pam_faillock

Issue #181, opened 2019-09-16
sssd: add `with-files-access-provider` to use SSSD's account management for local users


pbrezina:
Since SSSD is capable of handling local system users, there can exist configurations that combines local users with other authentication/access mechanisms. SSSD profile should support such configurations.

RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1734094

pbrezina:
* `master`
  * 6fca96ef70a88b138da22961dfaef67160e1ffbb - sssd: add `with-files-access-provider` to use SSSD's account management for local users

PR #190, opened 2020-01-14
sssd: add `with-files-access-provider` to use SSSD's account management for local users

pbrezina:
Because SSSD now handles local users it is a valid use case to call
account management in pam_sss for those users as well.

Adds new `without-files-provider` feature to sssd profile.

Resolves:
https://github.com/pbrezina/authselect/issues/181

### DIFF
--- a/profiles/sssd/README
+++ b/profiles/sssd/README
@@ -73,6 +73,16 @@ with-sudo::
 with-pamaccess::
     Check access.conf during account authorization.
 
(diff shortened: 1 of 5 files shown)


pbrezina:
* `master`
  * 6fca96ef70a88b138da22961dfaef67160e1ffbb - sssd: add `with-files-access-provider` to use SSSD's account management for local users

Issue #182, opened 2019-09-16
Read min uid from login.defs


pbrezina:
RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1745136

pbrezina:
This will be handled in new PAM module that will be then used by authselect.

pbrezina:
linux-pam pr: https://github.com/linux-pam/linux-pam/pull/166

pbrezina:
* `master`
  * 4cfb20423ca9a3e9b2b3de627b751e61fb0cf560 - profiles: switch uid checks to pam_usertype

PR #195, opened 2020-01-28
profiles: switch uid checks to pam_usertype

pbrezina:
`pam_usertype` is a new module that checks if it is a system or regular
user account. The checks respects settings from /etc/logins.defs so
the value is no longer hardcoded.

Resolves:
https://github.com/pbrezina/authselect/issues/182

### DIFF
--- a/profiles/nis/fingerprint-auth
+++ b/profiles/nis/fingerprint-auth
@@ -9,7 +9,7 @@ account     required                                     pam_access.so          
 account     required                                     pam_faillock.so                                        {include if "with-faillock"}
 account     required                                     pam_unix.so broken_shadow
 account     sufficient                                   pam_localuser.so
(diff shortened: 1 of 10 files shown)


pbrezina:
Because this is a new module, we need to also update pam dependency. I'm waiting for the module to be present in Fedora.

pbrezina:
Thank you. I opened https://github.com/pbrezina/authselect/issues/203 to not forget about spec file change once pam_usertype lands in Fedora.

pbrezina:
* `master`
  * accd7e554e3b16faf4192c84b1216bf30db2de9e - nis: simplify account phase
  * 85488ddaf38f06efc5db881778c6a56dd0092e78 - profiles: use pam_faillock also for system accounts
  * 4cfb20423ca9a3e9b2b3de627b751e61fb0cf560 - profiles: switch uid checks to pam_usertype

PR #202, opened 2020-02-12
fix backup-restore of non-authselect configuration

pbrezina:
`backup-restore` did not work correctly when `dconf` files were missing

This PR also contains further improvements, see commit messages for details.

### DIFF
--- a/src/lib/util/selinux.c
+++ b/src/lib/util/selinux.c
@@ -369,6 +369,12 @@ selinux_copy_files_safely(struct selinux_safe_copy *table,
     /* First, write content into temporary files, so we can safely fail
      * on error without overwriting destination files. */
     for (i = 0; table[i].source != NULL; i++) {
+        if (file_exists(table[i].source) == ENOENT) {
+            INFO("File [%s] does not exist", table[i].source);
+            tmpfiles[i] = NULL;
+            continue;
+        }
+
         INFO("Writing temporary file for [%s]", table[i].destination);
         ret = selinux_mkstemp_copy(table[i].source, dirs[i], names[i],
                                    dir_mode, &tmpfiles[i]);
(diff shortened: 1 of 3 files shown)


src/lib/util/selinux.c
```diff
@@ -369,6 +369,18 @@ selinux_copy_files_safely(struct selinux_safe_copy *table,
     /* First, write content into temporary files, so we can safely fail
      * on error without overwriting destination files. */
     for (i = 0; table[i].source != NULL; i++) {
+        if (file_exists(table[i].source) == ENOENT) {
+            if (!table[i].can_unlink) {
+                ERROR("File [%s] does not exist but it is not safe to "
+                      "unlink [%s].", table[i].source, table[i].destination);
```
thalman: I understand this error, but for administrators the message may look strange.
Why deleting is not safe when the file is actually gone? Also I would replace s/unlink/remove or delete/.

Maybe something like "File [%s] does not exist in selected profile but it is not safe to delete [%s]."

thalman: I realized that with full path expanded it is a bit clearer. If you think that the error message is better as it is, let me know.

pbrezina: How about now? This is really just a corner case if someone deleted the file from backup. It should not happen.

pbrezina:
* `master`
  * c8fa7098fc46cf583ccd0ad37792dc9f6509f48e - backup-list: return empty array if directory does not exist
  * 59149a3487e0487c8d4138d000de9ed9afeef1a8 - backup: do not unlink dangerous missing files
  * 8a3ea7d955392c76e4f62e1efb23dcfd930c3f6e - backup: remove authselect.conf when restoring non-authselect configuration
  * 5165e5e62672f61498d3c3fbc77897171d1f9273 - backup: remove non-existing files when restoring non-authselect configuration

Issue #203, opened 2020-02-17
spec: add dependency on pam package that contains pam_usertype


pbrezina:
https://github.com/pbrezina/authselect/issues/182 added a new pam module `pam_usertype`. It is already in https://github.com/linux-pam/linux-pam but not yet in Fedora. We should add requirements to spec file once it lands in Fedora.

PR #204, opened 2020-03-04
spec: add dependency on pam that contains pam_usertype.so

pbrezina:
With this, we are dropping support of releases older than Fedora 32.

Resolves:
https://github.com/pbrezina/authselect/issues/203


pbrezina:
* `master`
  * 41b7f732eb30e790746e0c2966bad920b8cf3398 - spec: add dependency on pam that contains pam_usertype.so

PR #205, opened 2020-03-06
new option: with-session-pam-env

jamescassell:
pam_env is useful for setting environment variables for all pam
sessions.  Among other things, this can be used to set http_proxy
environment variables on a system-wide basis to allow normal networking
functionality on networks that can access the internet only via a proxy.

The existing pam_env in the auth section does not have any effect for
pam sessions that bypass the auth modules, such as cached sudo and
publickey SSH login.

with-session-pam-env enables the pam_env session module in addition to
the always-enabled auth module.  This makes /etc/security/pam_env.conf
settings also take effect for public-key-authenticated SSH and cached
sudo.

### DIFF
--- a/profiles/minimal/README
+++ b/profiles/minimal/README
@@ -35,6 +35,10 @@ with-pamaccess::
 with-altfiles::
     Use nss_altfiles for passwd and group nsswitch databases.
 
+with-session-pam-env::
+    Enable pam_env sessions module.  Useful for setting environment variables
+    for all users via /etc/security/pam_env.conf.
+
 without-nullok::
     Do not add nullok parameter to pam_unix.
 
--- a/profiles/minimal/password-auth
+++ b/profiles/minimal/password-auth
@@ -13,6 +13,7 @@ password    requisite                                    pam_pwquality.so try_fi
 password    sufficient                                   pam_unix.so sha512 shadow {if not "without-nullok":nullok} try_first_pass use_authtok
 password    required                                     pam_deny.so
 
+session     optional                                     pam_env.so                                            {include if "with-session-pam-env"}
 session     optional                                     pam_keyinit.so revoke
 session     required                                     pam_limits.so
 session     optional                                     pam_ecryptfs.so unwrap                                {include if "with-ecryptfs"}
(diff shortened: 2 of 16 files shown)


jamescassell:
Is this a good change, a bad change?  Is the an IRC channel I should join to chat about it? An e-mail list? A web forum?  A Red Hat Support Case?

pbrezina:
Hi I'm sorry, I totally missed this PR. Thank you for the reminding me. Patch looks good to me from a quick look but I need to check the use case.

t8m:
There might be unexpected results for applications that call both pam_setcred and pam_open_session effectively calling pam_env twice. I believe the correct thing to do would be rather to fix the applications that do not call pam_setcred properly to call it.

jamescassell:
> There might be unexpected results for applications that call both pam_setcred and pam_open_session effectively calling pam_env twice. I believe the correct thing to do would be rather to fix the applications that do not call pam_setcred properly to call it.
> 

Are you suggesting to have this option also remove the auth module? Several paths skip the auth module, including ssh via public key, sudo NOPASSWD, and cached sudo where a password was required for the first invocation but not subsequent. IIRC, cron also skips the auth module.

t8m:
Please note that not calling pam_authenticate() does not mean automatically not calling pam_setcred().

ssh via public key should not skip the pam_setcred call (if it does, it would be a bug in sshd).
cronie crond does not skip the pam_setcred call.
sudo should not skip the pam_setcred call (if it does, it would be a bug in sudo).

pbrezina:
How does `pam_setcred` invoke `pam_env` from auth?

t8m:
`auth required pam_env.so` means that not only pam_env.so will be called from pam_authenticate() call but also from pam_setcred() call even if the calling application does not call pam_authenticate() at all.

pbrezina:
Ah, I see. `pam_env` is actually called through [`pam_setcred()`](https://github.com/linux-pam/linux-pam/blob/master/modules/pam_env/pam_env.c#L872) rather then [`pam_authenticate()`](https://github.com/linux-pam/linux-pam/blob/master/modules/pam_env/pam_env.c#L796).

Thank you for the PR but unfortunately I'm going to close it per Tomas' comments. Please, open bug on the problematic applications to rather call `pam_setcred()`.

PR #208, opened 2020-05-05
file_copy: use fchmod() and pass only valid mode bits

mgerstner:
Currently in `strace` authselect shows strange system calls like:

chmod("/etc/pam.d/system-auth.NenLC7", 0100644)

The mode parameter actually also contains the type bits, because the
`st_mode` member that was previously obtained from `fstat()` is passed
unmodified.

In this commit a bitwise AND is performed to restrict the parameter to
the permission bits only. Furthermore the `chmod` is changed into an
`fchmod` which is more robust and faster in this context.

### DIFF
--- a/src/lib/util/file.c
+++ b/src/lib/util/file.c
@@ -513,7 +513,7 @@ file_copy(const char *source,
 
     /* Restore original owner and mode.  Errors here are not fatal, since we
      * have the original content already stored and owned by root. */
-    ret = chmod(destpath, statbuf.st_mode);
+    ret = fchmod(fileno(fdest), statbuf.st_mode & ALLPERMS);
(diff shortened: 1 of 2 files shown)


pbrezina:
Nice catch, thank you. Can you also switch to `fchmod` at https://github.com/authselect/authselect/blob/master/src/lib/util/textfile.c#L172 please?

pbrezina:
Thank you. I opened an issue for it: #211 

* `master`
  * d21adb9c953888248bf163e4b9d0808253c7a873 - textfile_write: use fchmod instead of chmod
  * a5b4adb51156c7361726c2afaeacfc0a8845e250 - file_copy: use fchmod() and pass only valid mode bits

PR #219, opened 2020-07-13
utils: correctly remove the whole line if `str` does not point to its beginning

pbrezina:
The following scenario triggers a bug when the line is not removed completely
and the two lines are merged instead.
```
BEGINNING {if "condition":true|false} END {include if "condition"}
NEXT LINE

->
BEGINNING falseNEXT LINE
```

This is because `match_string` points after the first condition and we
only remove the line to this point. Therefore we need to interate
before `match_string` so we can find the real line start.

Resolves:
https://github.com/authselect/authselect/issues/218


thalman:
Thank you for the patch, ACK

pbrezina:
* `master`
  * 4b913748e413314b69c315c314c3d07e10471712

PR #230, opened 2020-11-18
lib: do not create symlink to non-existent files when creating new profile

pbrezina:
Resolves:
https://github.com/authselect/authselect/issues/228


thalman:
Thank you for the patch, ACK

pbrezina:
* `master`
  * 02a085d1fbfc8c36a21730e547962e933f322263 - lib: do not create symlink to non-existent files when creating new profile

PR #239, opened 2021-03-09
fingerprint: Retain error code returned by pam_fprintd.so

benzea:
pam_fprintd.so will return PAM_AUTHINFO_UNAVAIL in some situations. It
is important that the fingerprint-auth stack returns this error code, so
that users (i.e. GDM) can tell the difference between an authentication
failure vs. an immediate return because no fingers are enrolled.

Fix this by changing sufficient to [success=ok default=bad] in order to
ensure that the pam_fprintd.so error code is returned.


pbrezina:
Thank you. Ack.

pbrezina:
* `master`
  * b1bd29712fe8fbcbd800cd7320d8851d0f7a436f - fingerprint: Retain error code returned by pam_fprintd.so

PR #242, opened 2021-03-25
fingerprint: Fix fingerprint authentication

benzea:
Commit b1bd29712fe8 ("fingerprint: Retain error code returned by
pam_fprintd.so") changed the pam_fprintd.so module to be included using
"required". This changes sets "default=bad", however changed the
"success" configuration from "success=done" to "success=ok" resulting in
the later pam_deny.so to prevent authentication.

The fix is simple, explicitly set "[success=done default=bad]" so that
we get the best of both worlds. Error reporting and working
authentication.


pbrezina:
Thank you. Ack.

pbrezina:
* `master`
  * df92cfa345ca7ab821a3ef873f6c4e5cdab76390 - fingerprint: Fix fingerprint authentication

PR #244, opened 2021-04-16
Add kde to the list of smartcard capable PAM services

### LINKED ISSUE #234: Need to add pam service name for plasma screen locker
The KDE 5/Plasma desktop uses the pam service name of "kde" for the screen lock.  I think this just need to be added to https://github.com/authselect/authselect/blob/master/profiles/sssd/system-auth#L5 but I'm not sure.

opoplawski:
Resolves #234

### DIFF
--- a/profiles/sssd/system-auth
+++ b/profiles/sssd/system-auth
@@ -2,7 +2,7 @@
 auth        required                                     pam_env.so
 auth        required                                     pam_faildelay.so delay=2000000
 auth        required                                     pam_faillock.so preauth silent                         {include if "with-faillock"}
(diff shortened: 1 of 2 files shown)


sumit-bose:
Hi,

I agree with adding "kde" to the list of the PAM services which can handle local Smartcard authentication. But I think it is not expected to add the change to all the `*.po` files. It should be sufficient to add the change `authselect-profiles.5.adoc` and the `*.po` files are updated by Makefile magic. @pbrezina, am I right here?

bye,
Sumit

pbrezina:
Yes, you are right. @opoplawski Thank you for your patch, could you please remove the changes in *po files?

opoplawski:
Done.  I assuming that included the .pot file as well.

pbrezina:
Thank you, ack.

PR #248, opened 2021-05-13
profiles: try_first_pass has no effect on pam_unix and pam_pwquality

pbrezina:
Resolves:
https://github.com/authselect/authselect/issues/247


ikerexxe:
I think that the man pages should also be changed to avoid confusion.

Apart from that everything looks fine to me.

pbrezina:
Makes sense. See new patch.

ikerexxe:
LGTM!

thalman:
ACK

PR #250, opened 2021-05-13
profiles: add pam_debug to empty fingerprint and smartcard stack

pbrezina:
This way if someone tries to use the stack it will return AUTHINFO_UNAVAIL.

   Resolves: https://github.com/authselect/authselect/issues/238


thalman:
Thank you for the patch, ACK

PR #255, opened 2021-06-18
sssd: default to files first for users and groups

pbrezina:
The passwd and group databases will now default to files first.
The order "sss files" can be enabled with "with-files-provider"
feature.

This relates to F35 change:
https://fedoraproject.org/wiki/Changes/FlexibleLocalUserCache

Resolves:
https://github.com/authselect/authselect/issues/254


sumit-bose:
Hi,

thank you, patch is looking good and is working as expected. ACK

bye,
Sumit

PR #256, opened 2021-06-29
cli: use gettext on common options

pbrezina:
Also make --debug description the same as in cli_tool_print_common_opts.

These options are printed when a wrong argument is given on the command line. E.g.
  authselect select --invalid-arg


thalman:
Thank you for the patch, ACK

PR #262, opened 2021-07-14
spec: remove authselect-compat package

pbrezina:
Resolves:
https://github.com/authselect/authselect/issues/261


thalman:
Looks good to me, thank you for the patch

PR #266, opened 2021-07-22
lib: avoid freeing uninitialized variable in authselect_apply_changes()

pbrezina:
If authselect_profile() fails, we goto done and try to free uninitialized
variable.

Resolves:
https://github.com/authselect/authselect/issues/265


thalman:
Thank you for the patch, ACK

PR #268, opened 2021-09-02
Update authselect.spec.in

jprostko:
spec: Add nss-altfiles for Kinoite

Like on Silverblue, this module is required on Kinoite as well.
This change extends the change in https://github.com/authselect/authselect/commit/9e14ff3cd1cb8a84561d1b8adeaf00c42c4ce5c4 to include Kinoite.

### DIFF
--- a/rpm/authselect.spec.in
+++ b/rpm/authselect.spec.in
@@ -247,8 +247,8 @@ if [ -f %{validfile} ]; then
     %__rm -f %{validfile}
 fi
 
-# Add nss-altfiles if we are on Silverblue
-if %__grep -i silverblue /etc/os-release &> /dev/null; then
+# Add nss-altfiles if we are on Kinoite or Silverblue
+if %__grep -Ei "kinoite|silverblue" /etc/os-release &> /dev/null; then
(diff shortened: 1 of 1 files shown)


jprostko:
Kinoite does not reference Silverblue at all in its /etc/os-release file, hence the need for this change.  I performed an ipa-client-install today on Kinoite and could not get far into a graphical boot afterward since all of the altfiles references in /etc/nsswitch.conf were gone, and therefore sshd and a lot of other critical services could not start due to the service users not being found.

Example grep on /etc/os-release while on the Kinoite rawhide branch:

[jprostko@galago ~]$ grep -Ei "kinoite|silverblue" /etc/os-release
VERSION="Rawhide.20210829.n.0 (Kinoite Prerelease)"
PRETTY_NAME="Fedora Linux Rawhide.20210829.n.0 (Kinoite Prerelease)"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora-kinoite/"
VARIANT="Kinoite"
VARIANT_ID=kinoite

pbrezina:
Thank you. Ack.

PR #274, opened 2021-10-20
Fix potential flaws found by CodeQL

pbrezina:


thalman:
Thank you for the patch, LGTM.

PR #289, opened 2022-01-05
Two ostree/rpm-ostree compatibility patches

cgwalters:
Default to `--no-backup` if system is not initialized

If there's no configuration, then there's nothing to back up.
This means things "just work" with rpm-ostree, which makes
`/var` a read-only bind mount while scripts execute (to help
preserve user data).

In fact, the whole "backup configuration" thing is ideally long
term unnecessary on ostree systems, because ostree has the
capability to make *all* configuration changes transactional.

But for now this is really about making the existing code in
the `%posttrans` work.  This idea isn't ostree specific; I don't
see a reason to try to back up nonexistent configuration during an Anaconda
run for example either.

---

Don't try to write to `/var/lib/authselect` if it's read-only

I haven't fully grasped what authselect is trying to do with
the files in `/var/lib/authselect`.  This patch changes things
so that we skip trying to write to them if the path is read-only,
as will occur in `rpm-ostree compose tree` at build time.

I *think* these are backup copies of the files authselect uses to
detect "drift" from the files it wrote at the time of invocation.

If so, I think it's better to store them in `/etc/authselect/_pristine`
or something.  Or, add checksums as xattrs on the files.

In any case, this is now sufficient to get me a booting FCOS
system in concert with https://github.com/coreos/rpm-ostree/pull/3308

---

### DIFF
--- a/src/cli/main.c
+++ b/src/cli/main.c
@@ -115,6 +115,23 @@ parse_profile_options(struct cli_cmdline *cmdline,
     return EOK;
 }
 
+/* Check if there's an /etc/nsswitch.conf file.  If not,
+ * it means authselect must not have been invoked.
+ */
+static errno_t
+system_is_initialized(bool* initialized)
+{
+    struct stat stbuf;
+    *initialized = false;
+    if (stat (authselect_path_nsswitch (), &stbuf) < 0) {
+        if (errno == ENOENT)
+            return EOK;
+        return errno;
+    }
+    *initialized = true;
+    return EOK;
+}
+
 static errno_t
 perform_backup(int quiet,
                int backup,
@@ -153,6 +170,7 @@ static errno_t activate(struct cli_cmdline *cmdline)
     int quiet = 0;
     errno_t ret;
     int i;
+    bool initialized = false;
 
     struct poptOption options[] = {
         {"force", 'f', POPT_ARG_VAL, &enforce, 1, _("Enforce changes"), NULL },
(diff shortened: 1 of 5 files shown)


cgwalters:
To be clear, this change makes authselect not fail at build time, but it still doesn't seem to work at runtime, which I'm debugging.

cgwalters:
Now tested in concert with https://github.com/coreos/rpm-ostree/pull/3308 (though I didn't test a traditional yum-based system yet).

richiedaze:
> I haven't fully grasped what authselect is trying to do with the files in `/var/lib/authselect`. This patch changes things so that we skip trying to write to them if the path is read-only, as will occur in `rpm-ostree compose tree` at build time.
> 
> I _think_ these are backup copies of the files authselect uses to detect "drift" from the files it wrote at the time of invocation.

Authselect uses **/var/lib/authselect** to store a copy of the current generated **profile files**. Authselect compares them to the ones in **/etc/authselect** to detect manual edits during the authselect test "profile" command. It also stores the previous backup files in case something goes wrong so that you can revert them. 

Lets say I want **systemd-homed** edits on my pam stack. If I edit it the files manually (which is the wrong way), authselect would complain because both files didn't match the current profile. The proper way would be to create a custom default profile, edited the templates, generated the profile, so that both would match.

pbrezina:
Thank you for the PR, let's keep the discussion on the bugzilla first. And continue here once we settle on a solution.

pbrezina:
Closing this in favor of https://github.com/authselect/authselect/pull/291

PR #291, opened 2022-01-31
drop strict change detection and add opt-out command

pbrezina:
One of the main authselect features is to refuse to overwrite user changes
to the configuration. To do so, it used to validate files and directory
structure and also files contents.

This however allowed to have a mixed configuration on the system, that was
created by authselect but then modified by user. This is no longer possible
and the behavior has changed in a way that any user changes made to
configuration created by authselect are overwritten without requiring the
*--force* parameter (files contents are no longer validated).

This will make the configuration consistent - users either use authselect
or not, nothing in between. It also makes authselect work on ostree server
side, where it is not possible to write to /var


dustymabe:
One small addition of information to make things clearer:

> where it is not possible to write to /var

where it is not possible to write to /var using RPM scriptlets.

src/man/authselect.8.adoc
```diff
@@ -363,6 +363,7 @@ The *authselect* can return these exit codes:
 * 3: Current configuration is not valid, it was edited without authselect.
 * 4: System configuration must be overwritten to activate an authselect profile, --force parameter is needed.
 * 5: Executed command must be run as root.
+* 6: No authselect and no non-authselect configuration was detected.
```
spoore1, line 366: This message looks a little confusing to me.  From the commit message, I take this to mean that "no configuration" was found.  If that's correct, it may make more sense to simply say "No configuration was detected"

Or, is there another way to word "non-authselect configuration"?    "user" or "local" configuration maybe?

thalman:
The code looks good to me. Works as expected in my fedora.

Just the new command `opt-out` is not added to bash completion. You can easily add it like this:

    COMMANDS=(select apply-changes list list-features show requirements current
              check test enable-feature disable-feature create-profile
              backup-list backup-remove backup-restore opt-out)

pbrezina:
Fixed, thank you. You can see the diff [here](https://github.com/authselect/authselect/compare/fbfc5c74c7318d2907aad527075814c505cc88f9..f0b5dd404301b35e6ff01dec52d15e4cc56fc747). (Note `free(features)` change is from rebase to latest master, it is not part of this PR.

thalman:
Thank you for the patch, ACK

PR #298, opened 2022-03-10
Add switch for pam_oath (oathtoolkit)

jcpunk:
This adds way for the default profiles to add the oath-toolkit one time password scheme.

https://www.nongnu.org/oath-toolkit/pam_oath.html
https://koji.fedoraproject.org/koji/packageinfo?packageID=16336

They could probably use a bit of review as my understanding of pam is limited.

### DIFF
--- a/profiles/minimal/README
+++ b/profiles/minimal/README
@@ -26,6 +26,10 @@ with-mkhomedir::
 with-ecryptfs::
     Enable automatic per-user ecryptfs.
 
+with-oathtool::
+    Enable authentication with oathtool secrets through *pam_oath*.
+    User secrets expected in /etc/users.oath
+
 with-silent-lastlog::
     Do not produce pam_lastlog message during login.
 
--- a/profiles/minimal/REQUIREMENTS
+++ b/profiles/minimal/REQUIREMENTS
@@ -1,5 +1,7 @@
 - with-mkhomedir is selected, make sure pam_oddjob_mkhomedir module                       {include if "with-mkhomedir"}
   is present and oddjobd service is enabled and active                                    {include if "with-mkhomedir"}
   - systemctl enable --now oddjobd.service                                                {include if "with-mkhomedir"}
-                                                                                          {include if "with-mkhomedir"}
+                                                                                          {include if "with-altfiles"}
 - with-altfiles is selected, make sure nss_altfiles module is present                     {include if "with-altfiles"}
+                                                                                          {include if "with-oathtool"}
+- with-oathtool is selected, make sure /etc/users.oath is present                         {include if "with-oathtool"}
--- a/profiles/minimal/password-auth
+++ b/profiles/minimal/password-auth
@@ -2,6 +2,7 @@ auth        required                                     pam_env.so
 auth        required                                     pam_faildelay.so delay=2000000
 auth        required                                     pam_faillock.so preauth silent                         {include if "with-faillock"}
 auth        sufficient                                   pam_unix.so {if not "without-nullok":nullok}
+auth        [user_unknown=ignore success=done]           pam_oath.so usersfile=/etc/users.oath                  {include if "with-oathtool"}
 auth        required                                     pam_faillock.so authfail                               {include if "with-faillock"}
 auth        required                                     pam_deny.so
 
--- a/profiles/minimal/system-auth
+++ b/profiles/minimal/system-auth
@@ -2,6 +2,7 @@ auth        required                                     pam_env.so
 auth        required                                     pam_faildelay.so delay=2000000
 auth        required                                     pam_faillock.so preauth silent                         {include if "with-faillock"}
 auth        sufficient                                   pam_unix.so {if not "without-nullok":nullok}
+auth        [user_unknown=ignore success=done]           pam_oath.so usersfile=/etc/users.oath                  {include if "with-oathtool"}
 auth        required                                     pam_faillock.so authfail                               {include if "with-faillock"}
 auth        required                                     pam_deny.so
 
--- a/profiles/nis/README
+++ b/profiles/nis/README
@@ -24,6 +24,10 @@ with-mkhomedir::
 with-ecryptfs::
     Enable automatic per-user ecryptfs.
 
+with-oathtool::
+    Enable authentication with oathtool secrets through *pam_oath*.
+    User secrets expected in /etc/users.oath
+
 with-fingerprint::
     Enable authentication with fingerprint reader through *pam_fprintd*.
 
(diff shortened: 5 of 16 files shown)


pbrezina:
Hi, thank you for your contribution. I'd like to get more information about the module first.

It is not clear to me how the user is expected to configure OTP generator? Also it looks like you are supposed to provide password in plain text in configuration file to enable 2FA, that does not look very secure.

Have you considered pam_u2f instead?

jcpunk:
I looked at `pam_u2f`, but hardware and software tokens are a bit tricky in my use case (bottom of a mine, 45m one way elevator ride).  Having something I can be sure I can relay to folks underground makes things tricky....

thmo:
IIUC, the scope of the two plugins is slightly different, e.g., `pam_oath` provides support for TOTP, which `pam_u2f` doesn't.

According to the [documentation](https://www.nongnu.org/oath-toolkit/pam_oath.html), it should be possible to configure `pam_oath` on a per-user base, using `$HOME` and `$USER` for the `usersfile` config.

jcpunk:
I wonder if it might make more sense to add a substack I could use via inheritance... in theory that would let me keep access to any features/fixes, but wouldn't become a never ending list of "yet another pam module" to maintain?

pbrezina:
> I wonder if it might make more sense to add a substack I could use via inheritance... in theory that would let me keep access to any features/fixes, but wouldn't become a never ending list of "yet another pam module" to maintain?

That is rather problematic because it is very easily to break things or introduce security holes. Also it is not a universal solution as many modules requires more difficult configuration (for example pam_faillock).

Please understand that I am not against inclusion of this module. I'm just trying to get enough information in order to maintain it. So far I haven't got the answers I need, the discussion got sidetracked by me mentioning pam_u2f.

Please, answer the following question "It is not clear to me how the user is expected to configure OTP generator?" and concern "Also it looks like you are supposed to provide password in plain text in configuration file to enable 2FA, that does not look very secure.".

jcpunk:
The shared secret is stored in plain text within (based on this config) `/etc/oathtool/users`.

If that is a deal breaker, I understand and have no objections to closing this out.

The workflow is basically, an admin puts the right line in the config.  Something like
```
HOTP/T60/6   exampleuser - MySecret
```

An admin can then generate the one time password on any system where they know the secret, the clocks are close, and have the `oathtool` command.

```
oathtool -s 60s --totp MySecret -d 6
```

In this example the password is good for 60 seconds.

thmo:
> The shared secret is stored in plain text within (based on this config) `/etc/oathtool/users`.

As I said earlier, I think the secret can also be stored in the users' homes, if the `usersfile` option makes use of `$HOME` or `$USER`.

It is stored in plain text, but I don't think there is a way around this  - it is needed when generating the OTP anyway.

jcpunk:
I switched up `usersfile` to use a file with the substitutions you've mentioned.

pbrezina:
Thank you, I will try it soon and move on with this PR.

pbrezina:
Can `usersfile` be configured in some config file rather then in PAM stack?

pbrezina:
@thalman Can you please review this change?

thalman:
> @thalman Can you please review this change?

Sure, I will do it this week.

pbrezina:
@thalman Bump.

thalman:
Sorry for the delay. I looked on the PR and in my opinion it looks good and since we have pam_oath in distribution, it also make sense.

I will do few more tests and see then how to proceed.

thalman:
Hi @jcpunk

Today I made few tests and I would say that pam stack like this works unexpectedly (I was able to login with any/wrong password, just if the OTP part is good).

I guess we want the user to enter his password and as a second factor the OTP code. A'm I right?

Tom

jcpunk:
hmmmm, that is odd, I thought I'd set it to require the password and the token....

thalman:
> hmmmm, that is odd, I thought I'd set it to require the password and the token....

OK, I will try to do more tests and see what is the issue.

Thanks,
Tom

pbrezina:
@thalman Any progress?

thalman:
Sorry for the delay,

I tried sssd profile with this feature and tested it with su - <user> command.

### Pam stack work unexpectedly

The pam stack configured this way looks like:
    ...
    auth        sufficient                                   pam_unix.so nullok
    auth        [default=1 ignore=ignore success=ok]         pam_usertype.so isregular
    auth        sufficient                                   pam_sss.so forward_pass
    auth        [user_unknown=ignore success=done]           pam_oath.so usersfile=${HOME}/.${USER}.oath
    auth        required                                     pam_deny.so

@pbrezina @jcpunk, correct me if I'm wrong but since pam_unix (as well as pam_sss) is marked as sufficient, then pam_oath is not required to succeed. On the other hand, wrong password with the right OPT code let me in. 


### oath file in user's home is problematic

If the file is not present the pam module fails with 

    [pam_oath.c:pam_sm_authenticate(488)] done. [Authentication failure]

It looks that if the admin enables the oath module, he must create the file[s] immediately for all users (empty file would do) otherwise they won't be able to log in.

If the file exists but the user is not in the file, then the oath module works fine and as expected (returns unknown user).

So I think one central file is actually better.

pbrezina:
> @pbrezina @jcpunk, correct me if I'm wrong but since pam_unix (as well as pam_sss) is marked as sufficient, then pam_oath is not required to succeed.

This is correct.

pbrezina:
Any update on this? I understand we took our time to review this, however we certainly do not want to discourage you from the contribution, I am sorry if you got this feeling.

thalman:
> > It looks that if the admin enables the oath module, he must create the file[s] immediately for all users (empty file would do) otherwise they won't be able to log in.
> 
> That feels like a sufficiently large blocker to close out this this without merging.

I would say it is not a blocker if we use one central file, like you did originally

    auth [user_unknown=ignore success=done] pam_oath.so usersfile=/etc/users.oath  {include if "with-oathtool"}

But the tricky part is to make this mandatory for configured users. With suggested patch it does not matter what you type in as an OTP code. It is enough to have the password right and the user is logged in.

SSH might be a bit special case - I do not know how exactly it evaluates the PAM stack but if you try `su - user` you can see it.

To make that work we have to somehow handle the lines that are currently `sufficient` (skip them and/or replace with another?).

I'm not sure what you want to achieve exactly (should that work for all users including those centrally managed in ldap/AD? Should that work for all services including GDM?).

jcpunk:
My primary target is local console access for local users.  In an ideal world any known user and any pam service would be nice.  I'm not sure what workflows other folks might have in mind.

Since two facter doesn't seem to be super likely, I can live with the fall back to OATH iff the password doesn't match.

thalman:
Hello @jspunk.

Thank you for the effort and I'm sorry that you PR is here for so long unattended. I consider this proposal interesting, but at the current state it is not working reliably and I did not have time to improve it on my own.

So after some hesitation I decided to close this PR. Feel free to reopen it, if you fix the problems mentioned in the conversation.

sincerely 
Tomáš

PR #301, opened 2022-03-30
sssd: Allow users unknown to sssd to authenticate

opoplawski:
On EL8 with:
```
Profile ID: sssd
Enabled features:
- with-smartcard
- with-smartcard-lock-on-removal
- with-sudo
- with-custom-automount
```
A local user (tempmon uid 1000 in /etc/passwd) cannot access cron:
```
You (tempmon) are not allowed to access to (crontab) because of pam configuration.
```
pam_sssd:
```
[pam_reply] (0x0200): Returning [10]: User not known to the underlying authentication module to the client [CID #2872]
```
I'm not entirely sure that this is the proper fix, but it seems likely.


pbrezina (approved):

EmptyByte:
Hi, did you fix it also for `with-smartcard-required` ? This Bug was fixed for `with-smartcard` but not `with-smartcard-required`. [BZ 2070325](https://bugzilla.redhat.com/show_bug.cgi?id=2070325) EL9 and EL8

PR #303, opened 2022-03-31
lib: correctly restore authselect backup

pbrezina:
The restore process did not write configuration copy to
/var/lib/authselect which made the restoration incomplete
and `authselect check` then failed.

Steps to reproduce:
1. authselect select sssd --force
2. authselect enable-feature with-mkhomedir
3. authselect apply-changes --backup=test
4. authselect backup-restore test
5. authselect check

`authselect check` will fail without this patch.

Resolves: https://github.com/authselect/authselect/issues/302


thalman (approved):
Thank you for the patch ACK

PR #311, opened 2022-06-03
Homed flags

alexpattyn:
Add a flag and pam config that allows homed created users to login via the CLI and graphical login screen.

### DIFF
--- a/profiles/sssd/REQUIREMENTS
+++ b/profiles/sssd/REQUIREMENTS
@@ -25,3 +25,6 @@ Make sure that SSSD service is configured and enabled. See SSSD documentation fo
 - with-gssapi is selected, make sure that GSSAPI authenticaiton is enabled in SSSD        {include if "with-gssapi"}
   - set pam_gssapi_services to a list of allowed services in /etc/sssd/sssd.conf          {include if "with-gssapi"}
   - see additional information in pam_sss_gss(8)                                          {include if "with-gssapi"}
+
+- with-homed is selected, make sure that the system-homed service is enabled              {include if "with-homed"}
+  - systemctl enable --now systemd-homed.service                                          {include if "with-homed"}
--- a/profiles/sssd/password-auth
+++ b/profiles/sssd/password-auth
@@ -7,6 +7,7 @@ auth        required                                     pam_u2f.so cue {if not 
 auth        [default=1 ignore=ignore success=ok]         pam_usertype.so isregular
 auth        [default=1 ignore=ignore success=ok]         pam_localuser.so
 auth        sufficient                                   pam_unix.so {if not "without-nullok":nullok}
+-auth       sufficient                                   pam_systemd_home.so                                    {include if "with-systemd-homed"}
 auth        [default=1 ignore=ignore success=ok]         pam_usertype.so isregular
 auth        sufficient                                   pam_sss.so forward_pass
 auth        required                                     pam_faillock.so authfail                               {include if "with-faillock"}
@@ -15,12 +16,14 @@ auth        required                                     pam_deny.so
 
 account     required                                     pam_access.so                                          {include if "with-pamaccess"}
 account     required                                     pam_faillock.so                                        {include if "with-faillock"}
+-account    sufficient                                   pam_systemd_home.so                                    {include if "with-systemd-homed"}
 account     required                                     pam_unix.so
 account     sufficient                                   pam_localuser.so                                       {exclude if "with-files-access-provider"}
 account     sufficient                                   pam_usertype.so issystem
 account     [default=bad success=ok user_unknown=ignore] pam_sss.so
 account     required                                     pam_permit.so
 
+-password   sufficient                                   pam_systemd_home.so                                    {include if "with-systemd-homed"}
 password    requisite                                    pam_pwquality.so local_users_only
 password    sufficient                                   pam_unix.so yescrypt shadow {if not "without-nullok":nullok} use_authtok
 password    sufficient                                   pam_sss.so use_authtok
@@ -28,9 +31,10 @@ password    required                                     pam_deny.so
 
 session     optional                                     pam_keyinit.so revoke
 session     required                                     pam_limits.so
-session     optional                                     pam_ecryptfs.so unwrap                                {include if "with-ecryptfs"}
+session     optional                                     pam_ecryptfs.so unwrap                                 {include if "with-ecryptfs"}
+-session    optional                                     pam_systemd_home.so                                    {include if "with-systemd-homed"}
 -session    optional                                     pam_systemd.so
-session     optional                                     pam_oddjob_mkhomedir.so                               {include if "with-mkhomedir"}
+session     optional                                     pam_oddjob_mkhomedir.so                                {include if "with-mkhomedir"}
 session     [success=1 default=ignore]                   pam_succeed_if.so service in crond quiet use_uid
 session     required                                     pam_unix.so
 session     optional                                     pam_sss.so
--- a/profiles/sssd/system-auth
+++ b/profiles/sssd/system-auth
@@ -12,6 +12,7 @@ auth        [default=1 ignore=ignore success=ok]         pam_localuser.so       
 auth        [default=2 ignore=ignore success=ok]         pam_localuser.so                                       {include if "with-smartcard"}
 auth        [success=done authinfo_unavail=ignore user_unknown=ignore ignore=ignore default=die] pam_sss.so try_cert_auth {include if "with-smartcard"}
 auth        sufficient                                   pam_unix.so {if not "without-nullok":nullok}
+-auth       sufficient                                   pam_systemd_home.so                                    {include if "with-systemd-homed"}
 auth        [default=1 ignore=ignore success=ok]         pam_usertype.so isregular                              {include if "with-gssapi"}
 auth        sufficient                                   pam_sss_gss.so                                         {include if "with-gssapi"}
 auth        [default=1 ignore=ignore success=ok]         pam_usertype.so isregular
@@ -22,12 +23,14 @@ auth        required                                     pam_deny.so
 
 account     required                                     pam_access.so                                          {include if "with-pamaccess"}
 account     required                                     pam_faillock.so                                        {include if "with-faillock"}
+-account    sufficient                                   pam_systemd_home.so                                    {include if "with-systemd-homed"}
 account     required                                     pam_unix.so
 account     sufficient                                   pam_localuser.so                                       {exclude if "with-files-access-provider"}
 account     sufficient                                   pam_usertype.so issystem
 account     [default=bad success=ok user_unknown=ignore] pam_sss.so
 account     required                                     pam_permit.so
 
+-password    sufficient                                  pam_systemd_home.so                                   {include if "with-systemd-homed"}
 password    requisite                                    pam_pwquality.so local_users_only
 password    sufficient                                   pam_unix.so yescrypt shadow {if not "without-nullok":nullok} use_authtok
 password    sufficient                                   pam_sss.so use_authtok
@@ -35,9 +38,10 @@ password    required                                     pam_deny.so
 
 session     optional                                     pam_keyinit.so revoke
 session     required                                     pam_limits.so
-session     optional                                     pam_ecryptfs.so unwrap                                {include if "with-ecryptfs"}
+session     optional                                     pam_ecryptfs.so unwrap                                 {include if "with-ecryptfs"}
+-session    optional                                     pam_systemd_home.so                                    {include if "with-systemd-homed"}
 -session    optional                                     pam_systemd.so
-session     optional                                     pam_oddjob_mkhomedir.so                               {include if "with-mkhomedir"}
+session     optional                                     pam_oddjob_mkhomedir.so                                {include if "with-mkhomedir"}
 session     [success=1 default=ignore]                   pam_succeed_if.so service in crond quiet use_uid
 session     required                                     pam_unix.so
 session     optional                                     pam_sss.so
--- a/profiles/winbind/REQUIREMENTS
+++ b/profiles/winbind/REQUIREMENTS
@@ -14,3 +14,6 @@ Make sure that winbind service is configured and enabled. See winbind documentat
 - with-mkhomedir is selected, make sure pam_oddjob_mkhomedir module                       {include if "with-mkhomedir"}
   is present and oddjobd service is enabled and active                                    {include if "with-mkhomedir"}
   - systemctl enable --now oddjobd.service                                                {include if "with-mkhomedir"}
+
+- with-homed is selected, make sure that the system-homed service is enabled              {include if "with-homed"}
+  - systemctl enable --now systemd-homed.service                                          {include if "with-homed"}
--- a/profiles/nis/README
+++ b/profiles/nis/README
@@ -62,6 +62,9 @@ with-mdns4::
 with-mdns6::
     Enable multicast DNS over IPv6.
 
+with-systemd-homed
+    If set, pam_systemd_homed is enabled for all pam operations.
+
 without-nullok::
     Do not add nullok parameter to pam_unix.
 
--- a/profiles/nis/REQUIREMENTS
+++ b/profiles/nis/REQUIREMENTS
@@ -14,3 +14,6 @@ Make sure that NIS service is configured and enabled. See NIS documentation for 
 - with-mkhomedir is selected, make sure pam_oddjob_mkhomedir module                       {include if "with-mkhomedir"}
   is present and oddjobd service is enabled and active                                    {include if "with-mkhomedir"}
   - systemctl enable --now oddjobd.service                                                {include if "with-mkhomedir"}
+
+- with-homed is selected, make sure that the system-homed service is enabled              {include if "with-homed"}
+  - systemctl enable --now systemd-homed.service                                          {include if "with-homed"}
--- a/profiles/nis/password-auth
+++ b/profiles/nis/password-auth
@@ -4,14 +4,17 @@ auth        required                                     pam_faillock.so preauth
 auth        sufficient                                   pam_u2f.so cue                                           {include if "with-pam-u2f"}
 auth        required                                     pam_u2f.so cue {if not "without-pam-u2f-nouserok":nouserok} {include if "with-pam-u2f-2fa"}
 auth        sufficient                                   pam_unix.so {if not "without-nullok":nullok}
+-auth       sufficient                                   pam_systemd_home.so                                    {include if "with-systemd-homed"}
 auth        required                                     pam_faillock.so authfail                                 {include if "with-faillock"}
 auth        optional                                     pam_gnome_keyring.so only_if=login auto_start            {include if "with-pam-gnome-keyring"}
 auth        required                                     pam_deny.so
 
 account     required                                     pam_access.so                                            {include if "with-pamaccess"}
 account     required                                     pam_faillock.so                                          {include if "with-faillock"}
+-account    sufficient                                   pam_systemd_home.so                                    {include if "with-systemd-homed"}
 account     required                                     pam_unix.so broken_shadow
 
+-password   sufficient                                   pam_systemd_home.so                                    {include if "with-systemd-homed"}
 password    requisite                                    pam_pwquality.so {if not "with-nispwquality":local_users_only}
 password    sufficient                                   pam_unix.so yescrypt shadow {if not "without-nullok":nullok} use_authtok nis
 password    required                                     pam_deny.so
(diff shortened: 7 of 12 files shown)


alexpattyn:
See https://github.com/authselect/authselect/issues/310

profiles/sssd/system-auth
```diff
@@ -19,6 +19,7 @@ auth        sufficient                                   pam_sss.so forward_pass
 auth        required                                     pam_faillock.so authfail                               {include if "with-faillock"}
 auth        optional                                     pam_gnome_keyring.so only_if=login auto_start          {include if "with-pam-gnome-keyring"}
 auth        required                                     pam_deny.so
+-auth       sufficient                                   pam_systemd_home.so                                    {include if "with-homed"}
```
rmader, line 22: For some reason this has to be above the `pam_deny.so` block in the `auth` section (also in the other file).

profiles/sssd/system-auth
```diff
@@ -27,11 +28,13 @@ account     sufficient                                   pam_localuser.so
 account     sufficient                                   pam_usertype.so issystem
 account     [default=bad success=ok user_unknown=ignore] pam_sss.so
 account     required                                     pam_permit.so
+-account    sufficient                                   pam_systemd_home.so                                    {include if "with-homed"}
```
rmader, line 31: The `pam_unix.so` apparently needs to be changed to `sufficient` to make this work - again I can't say why. So maybe something like the following would work:
```
account required pam_unix.so {exclude if "with-homed"}
account sufficient pam_unix.so {include if "with-homed"}
```

alexpattyn:
@rmader I didn't make it clear in this PR, but I am referencing https://www.freedesktop.org/software/systemd/man/pam_systemd_home.html for the changes I made.

But I am certainly not a PAM expert.

rmader:
> Yeah I see those now. Should we stick with the freedesktop docs then?

Maybe just wait for a review/opinion by an actual reviewer :)

alexpattyn:
> Maybe just wait for a review/opinion by an actual reviewer :)

Thought you were an actual reviewer haha But sounds good, thanks for checking the changes out!

profiles/sssd/password-auth
```diff
@@ -12,6 +12,7 @@ auth        sufficient                                   pam_sss.so forward_pass
 auth        required                                     pam_faillock.so authfail                               {include if "with-faillock"}
 auth        optional                                     pam_gnome_keyring.so auto_start                        {include if "with-pam-gnome-keyring"}
 auth        required                                     pam_deny.so
+-auth       sufficient                                   pam_systemd_home.so                                    {include if "with-homed"}
```
poettering, line 15: if you put this behin pam_deny, then this means pam-deny will already refuse the login attemp.

This should be placed right after the pam_unix.so line above i figure.

profiles/sssd/password-auth
```diff
@@ -20,11 +21,13 @@ account     sufficient                                   pam_localuser.so
 account     sufficient                                   pam_usertype.so issystem
 account     [default=bad success=ok user_unknown=ignore] pam_sss.so
 account     required                                     pam_permit.so
+-account    sufficient                                   pam_systemd_home.so                                    {include if "with-homed"}
```
poettering, line 24: pam_permit should remain last in this stack. 

Place this imediately before pam_unix here.

profiles/sssd/password-auth
```diff
@@ -20,11 +21,13 @@ account     sufficient                                   pam_localuser.so
...
 
 password    requisite                                    pam_pwquality.so local_users_only
 password    sufficient                                   pam_unix.so yescrypt shadow {if not "without-nullok":nullok} use_authtok
 password    sufficient                                   pam_sss.so use_authtok
 password    required                                     pam_deny.so
+-password   sufficient                                   pam_systemd_home.so                                    {include if "with-homed"}
```
poettering, line 30: here too, this makes no sense to put after pam_deny, as the request will then be refused already. this should be put first in this stack actually, since we don't want to run the rest of the stack at all for homed users (it does its own pwquality anyway, no need to repeat it)

pbrezina:
Hi, I'd like to learn little bit more from about the module.

From man page:
> Automatically mount home directories managed by systemd-homed.service on login, and unmount them on logout

I read this as it only mounts home directories but does not perform any authentication. In that sense, `sufficient` is wrong as it would allow user to login if the home directory is mounted so it should be required (but only for users that requires mounting their home dir).

There is Fedora BZ opened by Lennart:
https://bugzilla.redhat.com/show_bug.cgi?id=2085485

poettering:
> I read this as it only mounts home directories but does not perform any authentication. In that sense, `sufficient` is wrong as it would allow user to login if the home directory is mounted so it should be required (but only for users that requires mounting their home dir).

It also does auth. It authenticates the user against the encryption password used on the backing store (for example the LUKS passphrase of the LUKS volume that is the users home dir). It's a key concept actually: logging in with a password, fido key, pkcs11 key is what actually inlocks your data, and without having authenticated first, the data is safely locked away, impossible to access. i.e. quite different from traditional UNIX home dirs, which are not encrypted separately and there's no connection between the secrets used for login and that unlock the data on disk.

profiles/sssd/password-auth
alexpattyn: Done

profiles/sssd/password-auth
alexpattyn: Done

profiles/sssd/password-auth
alexpattyn: Done

profiles/sssd/password-auth
```diff
@@ -35,3 +38,4 @@ session     [success=1 default=ignore]                   pam_succeed_if.so servi
 session     required                                     pam_unix.so
 session     optional                                     pam_sss.so
 session     optional                                     pam_gnome_keyring.so auto_start                        {include if "with-pam-gnome-keyring"}
+-session    optional                                     pam_systemd_home.so                                    {include if "with-homed"}
```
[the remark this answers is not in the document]
alexpattyn: Done

alexpattyn:
Thanks for the review Lennart :+1:  Made the adjustments as requested.

rmader:
@alexpattyn: Thanks! The same changes need to be applied to `system-auth` - also, can you squash the fixing commits into the original ones (using `git rebase -i ...`)?

alexpattyn:
@rmader Okay everything should be squashed now.

profiles/sssd/system-auth
```diff
@@ -12,33 +12,41 @@ auth        [default=1 ignore=ignore success=ok]         pam_localuser.so
...
 session     required                                     pam_limits.so
 session     optional                                     pam_ecryptfs.so unwrap                                {include if "with-ecryptfs"}
+-session    optional                                     pam_systemd_home.so                                    {include if "with-homed"}
 -session    optional                                     pam_systemd.so
 session     optional                                     pam_oddjob_mkhomedir.so                               {include if "with-mkhomedir"}
 session     [success=1 default=ignore]                   pam_succeed_if.so service in crond quiet use_uid
 session     required                                     pam_unix.so
 session     optional                                     pam_sss.so
 session     optional                                     pam_gnome_keyring.so only_if=login auto_start          {include if "with-pam-gnome-keyring"}
+-session    optional                                     pam_systemd_home.so                                    {include if "with-homed"}
```
rmader, line 52: This ended up being duplicated now (also all other entries).

alexpattyn: Should be corrected now. Just have to squash again.

alexpattyn:
Already everything is squashed and corrected based on Lennart's comments.

profiles/sssd/password-auth
```diff
@@ -29,6 +32,7 @@ password    required                                     pam_deny.so
 session     optional                                     pam_keyinit.so revoke
 session     required                                     pam_limits.so
 session     optional                                     pam_ecryptfs.so unwrap                                {include if "with-ecryptfs"}
+-session    optional                                     pam_systemd_home.so                                    {include if "with-homed"}
```
poettering, line 35: indentation of the {} macro on the right is one character off

profiles/sssd/system-auth
```diff
@@ -36,6 +39,7 @@ password    required                                     pam_deny.so
 session     optional                                     pam_keyinit.so revoke
 session     required                                     pam_limits.so
 session     optional                                     pam_ecryptfs.so unwrap                                {include if "with-ecryptfs"}
+-session    optional                                     pam_systemd_home.so                                    {include if "with-homed"}
```
poettering, line 42: ditto

alexpattyn:
@pbrezina Let me know if further changes are needed :+1:

profiles/sssd/password-auth
```diff
@@ -15,22 +16,25 @@ auth        required                                     pam_deny.so
 
 account     required                                     pam_access.so                                          {include if "with-pamaccess"}
 account     required                                     pam_faillock.so                                        {include if "with-faillock"}
+-account    sufficient                                   pam_systemd_home.so                                    {include if "with-homed"}
```
pbrezina, line 19: @poettering Shouldn't this go after pam_unix?

profiles/sssd/password-auth
```diff
@@ -15,22 +16,25 @@ auth        required                                     pam_deny.so
...
 account     required                                     pam_unix.so
 account     sufficient                                   pam_localuser.so                                       {exclude if "with-files-access-provider"}
 account     sufficient                                   pam_usertype.so issystem
 account     [default=bad success=ok user_unknown=ignore] pam_sss.so
 account     required                                     pam_permit.so
 
+-password   sufficient                                   pam_systemd_home.so                                    {include if "with-homed"}
```
pbrezina, line 26: @poettering Shouldn't this go after pam_pwquality?

pbrezina (changes requested):
* Please, avoid changing anything else then pam_systemd_home to simplify the review and possible patch revert.
* Please, change commit message to `profiles: add with-homed` and include some description of the module.
* Some comments inside (applies for both password-auth and system-auth)
* I think you also want to include this in other profiles as well.

profiles/sssd/password-auth
poettering: No, before. homed links directly against libpwq and validates *all* passwords you set with it anyway, i.e. both when you initially create the user, and when you change the password later via PAM.

There's also a per-user setting that allows you to disable the password policy, if you like. I use that a lot when testing (the policy is just too frickin' annoying when you test passwords all the time). Since pam_pwquality can't really support that (as it has no idea of homed user records), it's hence kinda essential that validation is left to homed, which can check that property first, and skip validation if user record says so.

(the password policy ignore flag can only be set by root on the user record, so it's not a security issue).

Anyway, because of that we want to cut the pw change stack short: if pam_systemd_home.so owns the user, nothing else should run.

profiles/sssd/password-auth
poettering: so, homed synthesizes NSS records including shadow records from its native JSON user records, which include the account mgmt settings. These are more expressive than the NSS ones (a true superset), and can give more precise error messages. Hence it's kinda essential to run pam_systemd_home first, so that it can do its thing and then shortcut the PAM stack, so that the pam_unix.so stuff doesn't run.

If you'd run pam_unix first, then it would run the checks first, possibly output some stuff, and then pam_systemd_homed would check things again, possibly find more finegrained restrictions and output the same info, but more precise. So you'd end up with duplicate output, but one is more precise then the other.

hence, this order should be OK. There's no loss of validation checks, but more friendly output if done this way (i.e. less duplication and more precise info)

pbrezina:
Thank you for the explanation. Now, I think, we are settled on the format. I'll wait for @alexpattyn to address other points of my previous comment and we can push.

alexpattyn:
@pbrezina Sounds good. I'll make the changes you requested. Can you clarify the following point?

> I think you also want to include this in other profiles as well.

You mean to add these changes to other profile files under `profiles/sssd`? If so which ones?

pbrezina:
No, I mean to add it also to winbind and nis profiles.

sabedevops:
I think this is missing changes to the profile `README` and `REQUIREMENTS` such that it documents the `with-homed` feature, and describes its dependency on the systemd-homed service being enabled?

Also, I don't think we've modified the `nsswitch.conf` to add `nss-systemd` to `shadow` and `gshadow`? This might be a separate flag like `with-nss-systemd`

Something like this:
README:
```diff
118a119,124
> with-homed:
>     If set, pam_systemd_homed is enabled for all pam operations.
> 
> with-nss-systemd:
>     if set, the nss-systemd module is added for all compatible databases.
```
REQUIREMENTS:
```diff
27a28,29
> - with-homed is selected, make sure that the system-homed service is enabled              {include if "with-homed"}
>   - systemctl enable --now systemd-homed.service                                          {include if "with-homed"}
```
nsswitch.conf:
```diff
3c3
< shadow:     files
---
> shadow:     files {if "with-nss-systemd":systemd}
14c14
< gshadow:    files
---
> gshadow:    files {if "with-nss-systemd":systemd}
```

pbrezina:
> I think this is missing changes to the profile `README` and `REQUIREMENTS` such that it documents the `with-homed` feature, and describes its dependency on the systemd-homed service being enabled?

Right.

> Also, I don't think we've modified the `nsswitch.conf` to add `nss-systemd` to `shadow` and `gshadow`? This might be a separate flag like `with-nss-systemd`

Is this relevant for this PR? Or should it be part of different PR? Also nss-systemd is on by default for passwd and group, perhaps it should be on by default here as well?

> 
> Something like this: README:
> 
> ```diff
> 118a119,124
> > with-homed:
> >     If set, pam_systemd_homed is enabled for all pam operations.
> > 
> > with-nss-systemd:
> >     if set, the nss-systemd module is added for all compatible databases.
> ```
> 
> REQUIREMENTS:
> 
> ```diff
> 27a28,29
> > - with-homed is selected, make sure that the system-homed service is enabled              {include if "with-homed"}
> >   - systemctl enable --now systemd-homed.service                                          {include if "with-homed"}
> ```
> 
> nsswitch.conf:
> 
> ```diff
> 3c3
> < shadow:     files
> ---
> > shadow:     files {if "with-nss-systemd":systemd}
> 14c14
> < gshadow:    files
> ---
> > gshadow:    files {if "with-nss-systemd":systemd}
> ```

sabedevops:
I believe it is relevant (per the man page), since it is used by `systemd-homed.service` and `systemd-userdb.service`. To the extent they are unconditionally necessary or enabled by default, I think only @poettering can fill in that gap.

https://www.freedesktop.org/software/systemd/man/nss-systemd.html

My suggested conditional (`with-nss-systemd`) is probably not correct here, and should be changed to instead be either on by default, or some other conditional (like a part of `with-homed` or `with-systemd-userdb` or similar).

pbrezina:
@alexpattyn Thank you for your contribution. Are you willing to continue on this PR? There was just opened #322 that implements the same thing so we could move forward with the other PR if you don't have time to finish this.

poettering:
nss-systemd should really be enabled unconditionally on systemd systems. It hooks up a bunch of different things. homed is just one. systemd service DynamicUser= users are another. systemd-nspawn's userns ranges are another. If systemd is used at all, then nss-systemd should be in the stack.

This is quite different from pam_systemd_homed, which is homed-specific, and if you don't use homed then there's no point in having htat.

alexpattyn:
Hi @pbrezina I can continue. Sorry, been busy with other projects. What needs to be done? I saw there has been a lot of discussion around additional changes...

pbrezina:
I opened https://github.com/authselect/authselect/pull/323 for nsswitch change (systemd for shadow and gshadow).

@alexpattyn Reiterating the required changes:
- rename the option to `with-systemd-homed`
- include this in README and REQUIREMENTS as well (see [this comment](https://github.com/authselect/authselect/pull/311#issuecomment-1233152972), just disregard comments about with-nss-systemd)
- and then do the same for winbind and nis profiles

alexpattyn:
Thanks for the summary @pbrezina, I'll aim to knock this out over the weekend!

alexpattyn:
Sorry for the long delay. Just added the requested changes. Let me know if anything should be done.

pbrezina:
Thank you. I will resolve the conflict locally.

@poettering Should the pam_systemd_home module also be enabled for fingerprint and smartcard authentication or does it require password?

poettering:
> Thank you. I will resolve the conflict locally.
> 
> @poettering Should the pam_systemd_home module also be enabled for fingerprint and smartcard authentication or does it require password?

So, homed natively supports PKCS#11 (i.e. smartcard) auth anyway. If you have enrolled that against your homed account, then you can always use that for unlocking. 

It's not entirely clear to me how authconfig's smartcard/fingerprint stuff work I must say. But I figure it doesn't really apply to homed at all. Here's the thing:

1. as mentioned PKCS11 is supported natively, so replicating that in external PAM modules is wrong, conceptually.
2. More importantly: homed always uses the actual cryptographic authentication device to unlock the encrypted user home directory. i.e. for fido2 the secret for LUKS is directly acquired from FIDO and for PKCS#11 the same. This means you cannot mix and match PAM authentication modules with homed: if you attempt to use a secondary module (such as the fingerprint one) then it might authenticate something, but it's not good enough for homed since it won't be able to acquire any secret it could use for disk encryption from that.

I know that more recent biometric devices actually can generate a cryptographic quote on the fingerprint (along with some timestamp) which you then can ask the TPM to verify and unlock a secret for you, once we have that I think hooking fingerprint stuff up with homed would make a ton of sense, but then too it would be homed talking directly to the fingerprint layer, since it needs that cryptographic quote: pam would not be useful as a transport for that.

Hence: i don't think homed fits into PAM's/authconfig's idea of fingerprintf/smartcard logic at all. For smartcard that's not bad, because it can do this already natively, and for fingerprint stuff we have nothing comparable right now, but pam_fingerprint won't be it.

pbrezina:
> > Thank you. I will resolve the conflict locally.
> > @poettering Should the pam_systemd_home module also be enabled for fingerprint and smartcard authentication or does it require password?
> 
> So, homed natively supports PKCS#11 (i.e. smartcard) auth anyway. If you have enrolled that against your homed account, then you can always use that for unlocking.
> 
> It's not entirely clear to me how authconfig's smartcard/fingerprint stuff work I must say. But I figure it doesn't really apply to homed at all. Here's the thing:

smartcard-auth, fingerprint-auth and password-auth files are used by gdm to shortcut the authentication process (i.e. try sc, fp and password authentication at the same time), these files implement only the preferred authentication methods where system-auth combines everything together.

So if smartcard and fingerprint can not be used to unlock homed through pam then it does not make sense to include it in smartcard-auth and fingerprint-auth.

poettering:
Hmm, so gdm is running three PAM stacks in parallel and uses the one that succeeds first and aborts the other two? Interesting.

So I guess we could add some support to pam_systemd_home for that too, i.e. that you can tell it whether to accept only passwords, or only pkcs11/fido2. And then we could add pam_systemd_home in "only-password" mode to the first PAM stack, and in only-"pkcs11/fido" mode to the second one (and once we add fingerprint auth also to the third one). But that doesn't exist right now. Will add to TODO list.

(Do you know if gdm sets some PAM item or env var or so on the session object indicating for which mode it is setting up the session? or is that only communicated through PAM module params?)

pbrezina:
Thank you for patience, its finally pushed. I resolved conflicts, added few changes and pushed as:

* `master`
  * ab338c75631278d1188a1bfb1bb8e4942852dbb4 - profiles: add with-systemd-homed
  * 1b9b0b2715cdfac40b10da8cebf2229858911f74 - profiles: fix readme for with-systemd-homed and warn if the module is not installed

> (Do you know if gdm sets some PAM item or env var or so on the session object indicating for which mode it is setting up the session? or is that only communicated through PAM module params?)

AFAIK only module parameters. You could potentially use also name of the pam service.

poettering:
Thanks so much for working on this! Much appreciated! That I finally can stop locally munging authselect's files is fantastic

freddyw:
Good stuff, I just switched back to the default authselect profile 'with-systemd-homed' instead of my own custom made profile (Fedora 37)  and was able to login without issues. Awesome!

pbrezina:
Thank you for your feedback!

alexpattyn:
Thanks for your feedback and help getting this approved :smile: Looking forward to using homed on my system now!

martinpitt:
Excellent, this works very well. Many thanks! [Another ugly hack gone](https://github.com/martinpitt/ostree-pitti-workstation/commit/1e300ed7b03ea79e46b22f78f0bf39ec1ba28b2d) :tada:

profiles/sssd/REQUIREMENTS
```diff
@@ -25,3 +25,6 @@ Make sure that SSSD service is configured and enabled. See SSSD documentation fo
 - with-gssapi is selected, make sure that GSSAPI authenticaiton is enabled in SSSD        {include if "with-gssapi"}
   - set pam_gssapi_services to a list of allowed services in /etc/sssd/sssd.conf          {include if "with-gssapi"}
   - see additional information in pam_sss_gss(8)                                          {include if "with-gssapi"}
+
+- with-homed is selected, make sure that the system-homed service is enabled              {include if "with-homed"}
```
sabedevops, line 29: Should this have been updated to `with-systemd-homed` after the textual change to the feature name?

profiles/winbind/REQUIREMENTS
```diff
@@ -14,3 +14,6 @@ Make sure that winbind service is configured and enabled. See winbind documentat
 - with-mkhomedir is selected, make sure pam_oddjob_mkhomedir module                       {include if "with-mkhomedir"}
   is present and oddjobd service is enabled and active                                    {include if "with-mkhomedir"}
   - systemctl enable --now oddjobd.service                                                {include if "with-mkhomedir"}
+
+- with-homed is selected, make sure that the system-homed service is enabled              {include if "with-homed"}
```
sabedevops, line 18: Ditto here

PR #318, opened 2022-08-05
profiles: Add with-libvirt feature

andreabolognani:
The libvirt NSS plugins can be used to conveniently connect to VMs running on the system.

See [https://libvirt.org/nss.html](https://libvirt.org/nss.html) for additional information. Once this is merged, I will update that page to advertise authselect as the preferred mechanism to enable the feature.

This is my first time contributing to authselect, so please don't hesitate to point out any mistake I might have made :)

### DIFF
--- a/profiles/nis/README
+++ b/profiles/nis/README
@@ -65,6 +65,10 @@ with-mdns6::
 without-nullok::
     Do not add nullok parameter to pam_unix.
 
+with-libvirt::
+    Enable connecting to libvirt VMs using the hostname configured in the
+    guest OS or, as a fallback, their name.
+
 EXAMPLES
 --------
 * Enable NIS with no additional modules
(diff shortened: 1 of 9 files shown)


pbrezina:
@thalman Would you like to look at this?

thalman:
> @thalman Would you like to look at this?

Sure, I will take a look.

thalman:
@pbrezina, do you have any idea why missing tests were not executed? CodeQL and flake8 are scheduled actions but they are due date.

Since this is not code change, we are fine, I just wonder...

pbrezina:
> @pbrezina, do you have any idea why missing tests were not executed? CodeQL and flake8 are scheduled actions but they are due date.

It sometimes happens, I suspect some bug in GHA

pbrezina:
@andreabolognani Can you please just remove it from the minimal profile? This profile is not something we want to generally extend and promote.

andreabolognani:
> Can you please just remove it from the minimal profile? This profile is not something we want to generally extend and promote.

Sure thing!

PR #322, opened 2022-09-22
profiles: add "with-systemd-home" feature

richiedaze:
pam_systemd_home ensures that home directories managed by systemd-homed.service(8) are automatically activated (mounted) on user login, and are deactivated (unmounted) when the last session of the user ends.

Resolves:
https://bugzilla.redhat.com/show_bug.cgi?id=1806949

### DIFF
--- a/profiles/minimal/README
+++ b/profiles/minimal/README
@@ -44,6 +44,9 @@ with-mdns6::
 without-nullok::
     Do not add nullok parameter to pam_unix.
 
+with-systemd-home::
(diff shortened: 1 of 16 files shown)


pbrezina:
Thank you for your contribution. There is already https://github.com/authselect/authselect/pull/311 implementing this, so I will continue there unless the author drops it or becomes unresponsive.

pbrezina:
Thank you, the other pull request was merged. homed support is now available via "with-systemd-homed"

PR #323, opened 2022-09-26
profiles: add systemd to nsswitch shadow and gshadow

pbrezina:
nss_systemd also implements shadow and gshadow so it should be set
for these databases as well.

### DIFF
--- a/profiles/minimal/nsswitch.conf
+++ b/profiles/minimal/nsswitch.conf
@@ -1,7 +1,8 @@
 # In order of likelihood of use to accelerate lookup.
 passwd:     files {if "with-altfiles":altfiles }systemd
-shadow:     files
+shadow:     files systemd
 group:      files {if "with-altfiles":altfiles }systemd
+gshadow:    files systemd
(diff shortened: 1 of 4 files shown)


profiles/minimal/nsswitch.conf
```diff
@@ -1,7 +1,8 @@
 # In order of likelihood of use to accelerate lookup.
 passwd:     files {if "with-altfiles":altfiles }systemd
-shadow:     files
+shadow:     files systemd
 group:      files {if "with-altfiles":altfiles }systemd
```
poettering, line 4: hmm, this should really use `[SUCCESS=merge]` before `systemd`, so that that users provided by nss-systemd can also be part of groups defined by something else.

(but that's not really an issue with this PR i ugess, but a pre-existing issue.)

poettering: i am not entirely sure how this actually should look like in the end, because if you have both `files` and `altfiles` enabled, then systemd might want to merge group member lists into records returned by either, but `altfiles` might not be read to merge stuff into `files` output?

poettering:
looks excellent, thanks for looking into this. this PR looks fine, but the pre-existing merge group stuff really deserves another look.

poettering (approved):

profiles/minimal/nsswitch.conf
pbrezina: Hmm. I'll look into it, nobody asked for this yet though, so perhaps it is not that important.

pbrezina:
Thank you for your review.

PR #325, opened 2022-10-17
Fix installation of manpage

winterheart:
Fix errors when required command (find, basename, sed, install) can be found somewhere else; don't using system enviroment variable PATH with brokes normal command invocations.


thalman (commented):
Looks good to me. I is also consistent with the way we use find/mkdir/... commands in other scripts.

thalman (approved):

PR #328, opened 2022-10-21
Updated profiles/sssd/README to reflect actual state.

alexey-tikhonov:


profiles/sssd/README
```diff
@@ -1,26 +1,24 @@
-Enable SSSD for system authentication (also for local users only)
-=================================================================
+Enable SSSD for system authentication (optionally also for local users)
+=======================================================================
```
pbrezina, line 2: I would keep the original sentence. The reason is that this profile can be used for both SSSD users and local users regardless on the files provider configuration. We don't have any other profile that should be used for local users only mode.

profiles/sssd/README
```diff
@@ -1,26 +1,24 @@
...
-SSSD provides a set of daemons to manage access to remote directories
-and authentication mechanisms such as LDAP, Kerberos or FreeIPA. It provides
+SSSD provides a set of daemons to manage access to remote directories and
+authentication mechanisms such as LDAP, Kerberos, FreeIPA or AD. It provides
 an NSS and PAM interface toward the system and a pluggable backend system
 to connect to multiple different account sources.
 
 More information about SSSD can be found on its project page:
-https://pagure.io/SSSD/sssd
+https://sssd.io/
```
pbrezina, line 13: Remove trailing slash?

profiles/sssd/README
```diff
@@ -1,26 +1,24 @@
...
 
-By default, local users are served from SSSD rather then local files if SSSD
-is enabled (however they authenticate via pam_unix). This has a performance
-benefit since SSSD caches the files content in fast in-memory cache and thus
-reduces number of disk operations.
+If configured, NSS information for local users also can be served by SSSD
+(however they authenticate via pam_unix).
```
pbrezina, line 16: Please, describe also the default behavior.

`If configured, NSS information for local users also can be served by SSSD (however they authenticate via pam_unix).`

I believe "also" is placed incorrectly? -> If configured, NSS information for local users can **also** be served by SSSD (however they authenticate via pam_unix).

pbrezina (changes requested):
Thank you. Please, see minor comments inside.

profiles/sssd/README
alexey-tikhonov: Just deleted.

PR #342, opened 2023-06-16
Add `pam_krb5` as a fallback auth method for `sssd` profile.

jcpunk:
I'm engaged in long term conversation with the `sssd` folks about authentication features of `pam_krb5` not present in `sssd`.

I maintain `pam_krb5` in Fedora and EPEL.

Having a simple switch to enable this fallback would be helpful.

### DIFF
--- a/profiles/sssd/README
+++ b/profiles/sssd/README
@@ -68,6 +68,9 @@ without-pam-u2f-nouserok::
     authentication configured will not be able to log in *INCLUDING* root.
     Make sure you are able to log in before losing root privileges.
 
+with-pam_krb5::
+    Enable fallback to pam_krb5 if other authentication methods fail.
+
 with-silent-lastlog::
     Do not produce pam_lastlog message during login.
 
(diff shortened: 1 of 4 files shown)


pbrezina:
Hi, can you please point me to the conversations with SSSD folks? I'm primarily SSSD developer but I am not aware of any discussion about pam_krb5.

jcpunk:
I'm working through support case 02262181 which is supposed to be in contact with the SSSD folks regarding possible ideas for https://bugzilla.redhat.com/show_bug.cgi?id=1623624 / https://github.com/SSSD/sssd/issues/5293

Shared user accounts are always a mess and really ugly, but I've not found a solution for my nitch workflow another way.  I'd be very open to a chat on exactly what I've got going on if you're curious.

jcpunk:
Things on my side are still very much the same.

pbrezina:
Thank you. Do I understand it correctly that you need the ability to anyone with access to unlock session as soon as possible to get yours application up and running where the previous person left it in order to avoid potential disaster?

Also you want to have single home/environment but delegate access to other resources via Kerberos tickets?

jcpunk:
That is correct.

pbrezina:
Thank you. I'm going to close this PR, please keep it as custom profile. Let's continue with the support case.

PR #344, opened 2023-06-29
profiles: do not try to change password via sssd for local users

pbrezina:
Steps to reproduce:
1. Create local user and set passsword
2. Log in as the local user
3. Run passwd and provide wrong password as "Current password"

"Current password" prompt should be printed only once.

Resolves: https://github.com/authselect/authselect/issues/338


thalman (approved):
LGTM, Thank you for the patch

PR #345, opened 2023-08-02
Add switch to permit including external config

jcpunk:
This adds a place where I can trivially seed an external config file for pam.

Example usage, since the pam_oathtool patch didn't get merged, this will provide a place I can drop in my config without needing to fork the entire profile.

### DIFF
--- a/profiles/minimal/README
+++ b/profiles/minimal/README
@@ -47,6 +47,11 @@ with-mdns6::
 with-systemd-homed::
     If set, pam_systemd_homed is enabled for all pam operations.
 
+with-custom-pam-include::
+    If set pam will include a file called `custom-authselect`.
+    You are responsible to ensure the file exists and has safe
+    values for your environment.
+
 without-nullok::
     Do not add nullok parameter to pam_unix.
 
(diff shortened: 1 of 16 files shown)


pbrezina:
Hi, first things first, do you want this change in RHEL or in Fedora? Because if you need it in RHEL, you need to go through customer portal first, otherwise I can not include it in current releases anyway.

If I understand Tomas' comment correctly, the oath patch was not rejected out of unwillingness but because it was not generally functional.

I understand that there is some burden maintaining custom profile and we should come up with a way to improve this area (at the very least we can hint that the base profile has changed). What you propose is potentially insecure but interesting idea, however I don't think it can work well with the way PAM works, because modules must be placed on exact lines. So let's talk about this.

Let's say that you start using this change and it works for you for pam_oath. But then you'll want to include the pam_krb5 module to ask for (username, principal, password) combination - you will want it to be much earlier in the stack, otherwise you will provide (username, password, principal, password). Then you will need include file at the beginning. Then another user comes with his or her needs and asks for include file in the middle, then before pam_unix, then after pam_unix and so on... this it is not manageable. PAM just does not work this way. You really need to have control over the full stack and that is where custom profiles come in.

jcpunk:
Hmmm, this was my best guess at something that would be "extensible enough" but also minimally invasive...

pbrezina:
Closing per https://github.com/authselect/authselect/pull/345#issuecomment-1663817090

PR #349, opened 2023-08-24
tests: ported authselect tests from downstream

danlavu:
- adding authselect tests
- adding test files copied from sssd 
  - conftest.py
  - data/readme.md
  - pytest.ini
  - pyproject.toml
  - requirements.txt
  - setup.cfg

### DIFF
--- a/src/tests/system/polarion.yaml
+++ b/src/tests/system/polarion.yaml
@@ -1,0 +1,35 @@
+testcase:
+  required:
+    title:
+      transform:
+        pattern: "^(.*)$"
+        replace: "IDM-SSSD-TC: \\1"
+      validate: "IDM-SSSD-TC: (.+)"
+    setup:
+      format: pre
+    steps:
+    expectedresults:
+    customerscenario:
+    caseimportance:
+    requirement:
+  optional:
+    id:
+      default: "idm-sssd-tc::{{ item.id }}"
+    caseautomation:
+      default: "automated"
+    casecomponent:
+      default: "authselect"
+    status:
+      default: "approved"
+    subsystemteam:
+      default: "sst_idm_sssd"
+    upstream:
+      default: "yes"
+    automation_script:
+      default: "{{ tests_url }}/{{ item.location.file }}#L{{ item.location.line }}"
+    testtype:
+      default: "functional"
+    caselevel:
+      default: "system"
+    teardown:
+      format: pre
--- a/src/tests/system/pytest.ini
+++ b/src/tests/system/pytest.ini
@@ -1,0 +1,5 @@
+[pytest]
+addopts = --strict-markers
+testpaths = tests
+markers =
+ticket_tools = bz,gh,jira
--- a/src/tests/system/conftest.py
+++ b/src/tests/system/conftest.py
@@ -1,0 +1,20 @@
+# Configuration file for multihost tests.
+
+from __future__ import annotations
+
+from pytest_mh import MultihostPlugin
+from sssd_test_framework.config import SSSDMultihostConfig
+
+# Load additional plugins
+pytest_plugins = (
+    "pytest_mh",
+    "pytest_importance",
+    "pytest_ticket",
+    "pytest_tier",
+    "sssd_test_framework.fixtures",
+)
+
+
+def pytest_plugin_registered(plugin) -> None:
+    if isinstance(plugin, MultihostPlugin):
+        plugin.config_class = SSSDMultihostConfig
--- a/src/tests/system/mhc.yaml
+++ b/src/tests/system/mhc.yaml
@@ -1,0 +1,73 @@
+domains:
+- id: sssd
+  hosts:
+  - hostname: client.test
+    role: client
+    artifacts:
+    - /etc/sssd/*
+    - /var/log/sssd/*
+    - /var/lib/sss/db/*
+
+  - hostname: master.ldap.test
+    role: ldap
+    config:
+      binddn: cn=Directory Manager
+      bindpw: Secret123
+      client:
+        ldap_tls_reqcert: demand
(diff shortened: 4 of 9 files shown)


sidecontrol:
Continuing the conversation from the SSSD PR https://github.com/SSSD/sssd/pull/6890

@Iker > I didn't explain myself correctly. If possible, I'd like to run all test cases in all available distributions in CI. Is there anything in RHEL that does not allow us to do so?

The authselect package can no longer be uninstalled in Fedora. There is a hard dependency with sudo.  The one minimal test cases "test_authselect__when_removed_the_uninstall_is_clean", uninstalls authselect entirely. This will only work with older versions where the sudo dependency does not exist. I don't know how many customers need this, but we should still test and support it as it works this way. We can certainly tell them this is no longer supported and they need to change with the times too.

pbrezina:
RHEL8 and RHEL9 are stuck with `1.2.x` version, how do we want to handle this? Keep old tests for RHEL8 and 9 and start using mh tests for RHEL10? Or backport it also to 1.2.x?

pbrezina:
Is this ready for review? I think you said that you are still working on it.

sidecontrol:
These tests should work for 1.2, I thought we discussed major releases we will use git branches, and minor releases, use pytest markers. 

I need to finish the PAM PR before this can be merged.

pbrezina:
> These tests should work for 1.2, I thought we discussed major releases we will use git branches, and minor releases, use pytest markers.

If these will work for 1.2 then we can backport it to 1.2.x and run it for 8 and 9 from this branch. RHEL10 can use master.

> I need to finish the PAM PR before this can be merged.

Thanks, let me know when it is ready for review.

danlavu:
@pbrezina This is ready for review. 

Also, I do not have developer access to authselect, so I'm unable to assign this issue to myself.

pbrezina:
Hi, I invited you to the organization.

I am not sure if I will be able to review this before the Christmas break (I'll try). Can you please also make the tests run during PR CI? See SSSD's ci.yaml for reference.

pbrezina (changes requested):
Hi Dan,
please, include mhc.yaml, polarion.yaml .

All tests here are for authselect, so test_authselect.py and test_authselect prefix does not make sense. It would be better to organize the test differently eg by profile.

Included test may be organized as:
* test_sssd
* test_minimal
* test_distribution (eg test_authselect__when_removed_the_uninstall_is_clean)

Tests "*enable_and_disable*" are not sufficient. Presence of specific pam module is irrelevant, what matters is functionality. You must test that the functionality is working correctly.

test_authselect__when_removed_the_uninstall_is_clean - the problem is not RHEL vs Fedora but behavior change between authselect version. This test belongs only to 1.2.x branch.

danlavu:
@pbrezina  ack, changing a lot of things and removing the non-functional tests, and making the placeholder/sanity into functional before the next review.

danlavu:
@pbrezina This is ready for review now. It depends on the following PRs. 

https://github.com/SSSD/sssd-test-framework/pull/86
https://github.com/SSSD/sssd-test-framework/pull/85
https://github.com/SSSD/sssd-test-framework/pull/72

src/tests/system/polarion.yaml
```diff
@@ -0,0 +1,35 @@
+testcase:
+  required:
+    title:
+      transform:
+        pattern: "^(.*)$"
+        replace: "IDM-SSSD-TC: \\1"
```
pbrezina, line 6: Do we really use SSSD for authselect?

src/tests/system/tests/test_minimal.py
```diff
@@ -0,0 +1,31 @@
...
+
+import pytest
+from sssd_test_framework.roles.client import Client
+from sssd_test_framework.topology import KnownTopology
+
+
+@pytest.mark.importance("critical")
+@pytest.mark.ticket(bz=1654018)
+@pytest.mark.topology(KnownTopology.Client)
+def test_minimal__local_user_authentication(client: Client):
```
pbrezina, line 11: Minimal profile is no longer available in master, (RHEL-10) please,  replace it with local.

Keep minimal for 1.2.x branch that should be used for RHEL-8,9.

pbrezina (changes requested):
Thanks, this is much better and I think we can go with this for now.

In long term, however, I would like to rework it to be less sssd-like and more focused on authselect. We can not test winbind or nis stuff with the sssd framework. Lots of features are shared between profiles so we could use topology parametrization here, with the topology representing a profile. We are not that much interested in the sssd providers since we rely on the posix interface.

But its good enough for the moment. Thank you.

src/tests/system/polarion.yaml
danlavu: Yes, for example, I'll have to change the work item in polarion to match the test case name, but it's currently set to. 

IDM-SSSD-TC: authselect: Authselect SSSD profile enable and disable PAM silent lastlog

src/tests/system/tests/test_minimal.py
danlavu: I'll create another pull request for the 1.2.x branch after this one, these tests will apply to 1.2.x , so I think a backport will be fine?

src/tests/system/pytest.ini
```diff
@@ -0,0 +1,7 @@
+[pytest]
+addopts = --strict-markers
+testpaths = tests
+markers =
+    rhel: marks test for rhel only
+    fedora: marks test for fedora only
```
pbrezina, line 6: Unused.

pbrezina (changes requested):
Hi Dan, see the inline comment. Also please, rebase it properly without the merge commit and we can push. Thank you.

danlavu:
Done and done.

pbrezina (approved):

Issue #355, opened 2023-12-07
Automatically run authselect on first boot


poettering:
I am kinda interested in making regular Fedora systems just work if they come up with an empty /etc/. Such a system needs /etc/nsswitch.conf and so on configured, and I was wondering if you might be interested in making authselect useful to help with that.

So what I had in mind is that authselect.rpm shipping with a unit file like the following:

```
[Unit]
ConditionFirstBoot=yes
ConditionPathIsReadWrite=/etc
DefaultDependencies=no
After=systemd-remount-fs.service
Wants=first-boot-complete.target
Before=first-boot-complete.target sysinit.target
Conflicts=shutdown.target
Before=shutdown.target

[Service]
ExecStart=authselect apply-changes
```

This would regenerate /etc/nsswitch.conf and related files on first boot (and first boot only, which is what systemd calls a boot where /etc/ is empty). It has no effect on regular systems.

And there's a second change I'd like to see done: I'd like that authselect looks at a /usr/lib/authselect/authselect.conf or so if it finds that /etc/authselect/authselect.conf doesn't exist (and only then). That would allow images to ship the default config they want in /usr/, but if users do their own config in /etc/ (or if authselect is used on classic distros) then it would always take precedence.

The above is untested, but it's the rough idea. I can cook up some patches if you are OK with this. I am fully aware that Fedora as a whole has not adopted the goal of allowing the system to boot up with unpopulated /etc/, all this is a goal of mine. But maybe I can convince you to adopt this scheme anyway, given it would not alter behaviour of classic Fedora systems at all.

If you'd be OK with changes like this I can cook up some patches, just let me know. Thanks!

poettering:
Other distributions (in particular SUSE) are tracking progress of this work towards allowing systems to boot with empty /etc/ here btw:

https://github.com/uapi-group/specifications/issues/76

keszybz:
> authselect looks at a /usr/lib/authselect/authselect.conf or so if it finds that /etc/authselect/authselect.conf doesn't exist (and only then)

The same of course would have to apply to the rest of the config that is currently under `/etc/authselect/`.

A second step would be to move those config files to `/usr/lib/authselect/` by default.

poettering:
> The same of course would have to apply to the rest of the config that is currently under `/etc/authselect/`.

As I understand that's the config generated by authselect. It's output of the tool, not input, hence supporting it to be read from /usr/ would not be necessary.

poettering:
[...]

PR #356, opened 2023-12-07
authselect: add fallback config file in /usr/ if the main one in /etc…

poettering:
…/ doesn't exist

This implements one of the two ideas from https://github.com/authselect/authselect/issues/355

The mechanism is simple: when trying to read the configuration file and /etc/authselect/authselect.conf does not exist, we'll look for /usr/share/authselect/authselect.conf as a fallback configuration file. When writing/modifying we do not bother with the fallback file however, as any changes should only hit /etc/, never /usr/.

This useful in scenarios where /usr/ is immutable and /etc/ initially comes up empty: running authconfig like that will consume the fallback configuration, but then write out the local configuration in /etc/.

See: #355

### DIFF
--- a/src/lib/files/config.c
+++ b/src/lib/files/config.c
@@ -70,6 +70,11 @@ authselect_config_read(char **_profile_id,
     ret = textfile_read(PATH_CONFIG_FILE,
                         AUTHSELECT_FILE_SIZE_LIMIT,
                         &content);
+    if (ret == ENOENT) {
+        ret = textfile_read(PATH_CONFIG_FALLBACK_FILE,
+                            AUTHSELECT_FILE_SIZE_LIMIT,
+                            &content);
+    }
     if (ret != EOK) {
         return ret;
     }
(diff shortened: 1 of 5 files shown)


src/lib/files/config.c
```diff
@@ -70,6 +70,12 @@ authselect_config_read(char **_profile_id,
     ret = textfile_read(PATH_CONFIG_FILE,
                         AUTHSELECT_FILE_SIZE_LIMIT,
                         &content);
+    if (ret == ENOENT) {
+        ret = textfile_read(PATH_CONFIG_FALLBACK_FILE,
+                            AUTHSELECT_FILE_SIZE_LIMIT,
+                            &content);
+    }
+
```
keszybz, line 78: Unnecessary newline.

keszybz (approved):
LGTM.

poettering:
fixed the issue @keszybz found.

@pbrezina any chance we can get my three PRs merged? Would so love to see these settled? At least this one for example should be really uncontroversial, no?

pbrezina:
Hi, I'm sorry that it takes so long. My priorities are elsewhere at this moment and so far I am pretty much a sole maintainer so I can't delegate it yet. But I will try to find some time and look into this in following two weeks to get it done finally.

pbrezina:
This would break current `authselect opt-out`. I implemented little bit differently, this PR is replaced by https://github.com/authselect/authselect/pull/451

PR #357, opened 2023-12-07
authselect: automatically run authselect apply-changes on first boot

poettering:
This installs an authselect.service file that is pulled in by sysinit.target and automatically runs "authselect apply-changes" on first boot (i.e. when /etc/ is unpopulated), and applies any available configuration.

Due to the ConditionFirstBoot= and ConditionPathIsReadWrite=/etc this has zero effect on systems which do not come up in a first boot mode, or where /etc/ is immutable.

See: #355

### DIFF
--- a/systemd/Makefile.am
+++ b/systemd/Makefile.am
@@ -1,0 +1,10 @@
+systemdunitdir=@SYSTEMD_UNIT_DIR@
+
+dist_systemdunit_DATA = \
+	authselect.service \
+	$(NULL)
+
+# Hook this into sysinit.target by default
+install-exec-hook:
+	$(MKDIR_P) $(DESTDIR)/$(systemdunitdir)/sysinit.target.wants
+	$(LN_S) -f ../authselect.service $(DESTDIR)/$(systemdunitdir)/sysinit.target.wants/authselect.service
--- a/Makefile.am
+++ b/Makefile.am
@@ -6,6 +6,7 @@ SUBDIRS= \
     src/lib \
     src/cli \
     src/tests \
+    systemd \
     $(NULL)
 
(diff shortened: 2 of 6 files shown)


systemd/Makefile.am
```diff
@@ -0,0 +1,10 @@
+systemdunitdir=@SYSTEMD_UNIT_DIR@
+
+dist_systemdunit_DATA = \
+	authselect.service \
+	$(NULL)
+
+# Hook this into sysinit.target by default
+install-exec-hook:
+	$(MKDIR_P) $(DESTDIR)/$(systemdunitdir)/sysinit.target.wants
+	$(LN_S) -f ../authselect.service $(DESTDIR)/$(systemdunitdir)/sysinit.target.wants/authselect.service
```
keszybz, line 10: Why not define an `[Install]` section in the unit file and make this dynamic? Static enablement can be pretty annoying.

keszybz (commented):
Looks reasonable apart from one question.

systemd/Makefile.am
poettering: Dunno, the system is not really usable without PAM working, hence I doubt we should advertise this as something you can disable.

keszybz: Yeah, but it's possible that you want to generate the config in a different way, for example config mgmt, or not update automatically, etc. When the enablement is hardcoded like that, the only easy way to disable is to mask, which creates other problems.

pbrezina:
I went with [Install], disabled by default. If distro or image wants to use it, it needs to add preset file or enable it manually.

This PR is replaced with https://github.com/authselect/authselect/pull/451

PR #358, opened 2023-12-14
profiles: merge groups records with [SUCCESS=merge]

poettering:
Services such as systemd-homed would like to advertise users which are part of system groups, such as "wheel". That only works if glibc's [SUCCESS=merge] feature is used in nsswitch.conf, so that group records from multiple sources are merged.

This is documented here:

https://www.freedesktop.org/software/systemd/man/latest/nss-systemd.html#Configuration%20in%20/etc/nsswitch.conf

This hence adds [SUCCESS=merge] expressions to all NSS modules listed in the "groups" lines.

### DIFF
--- a/profiles/local/nsswitch.conf
+++ b/profiles/local/nsswitch.conf
@@ -1,7 +1,7 @@
 # In order of likelihood of use to accelerate lookup.
 passwd:     files {if "with-altfiles":altfiles }systemd
 shadow:     files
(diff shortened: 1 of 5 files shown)


pbrezina:
Actually, I did this for you. I also added required spec file change in separate commit.

poettering:
Excellent! thanks so much!

PR #368, opened 2024-02-28
spec: strengthen altfiles sed invocation

jlebon:
The regex that was given matched the `group` line but not the `passwd` line (which doesn't have that `[SUCCESS=merge]` bit).

Generalize the regex instead so that we just take whatever the contents of the conditional was and put it in there.

Fixes: https://github.com/coreos/fedora-coreos-tracker/issues/1680
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2266344

(Upstreamed from https://src.fedoraproject.org/rpms/authselect/pull-request/22)


pbrezina:
Thank you.

pbrezina (approved):

PR #377, opened 2024-06-28
spec: use Suggests(meta) to avoid circular dependencies

pbrezina:
See https://bugzilla.redhat.com/show_bug.cgi?id=2291235 for
more information.


pbrezina:
Copr still does not talk the new spec file meta, can't push yet.

PR #380, opened 2024-08-26
profiles: include "systemd" in shadow + gshadow nsswitch.conf lines

poettering:
Since systemd v249 (2021) nss-systemd hasn't just been generating entries in the "passwd" and "group" databases, but in "shadow" and "gshadow" as well. This is documented in the man page:

https://www.freedesktop.org/software/systemd/man/latest/nss-systemd.html

This is generally useful, since various account validity fields that userdb/homed maintains are only exported via spwd/sgrp, but most importantly this makes the "systemd-nspawn --bind-user=" logic work, which just propagates the hashed user pw into a container.

The patch is pretty simple: in all profiles where passwd/group already lists "systemd", we also add it to the lines with shadow/spwd, to the end.

### DIFF
--- a/profiles/local/nsswitch.conf
+++ b/profiles/local/nsswitch.conf
@@ -1,6 +1,6 @@
 # In order of likelihood of use to accelerate lookup.
 passwd:     files {if "with-altfiles":altfiles }systemd
-shadow:     files
+shadow:     files systemd
 group:      files [SUCCESS=merge] {if "with-altfiles":altfiles [SUCCESS=merge] }systemd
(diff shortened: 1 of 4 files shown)


keszybz (approved):
LGTM.

This is a natural followup to the previous patches that added `systemd` to `passwd:` and other lines.

pbrezina:
Hi, we already had this once: https://github.com/authselect/authselect/pull/323

Reverted in: 1e78f7e048747024a846fd22d68afc6993734e92

See: https://bugzilla.redhat.com/show_bug.cgi?id=2150155

Was this resolved?

pbrezina:
Test in Fedora CI with this PR applied is green now, so I suppose that the issue was resolved. Thank you.

PR #383, opened 2024-09-26
tests: adding misc test to resolve client hostname

danlavu:


danlavu:
I realized the fix was applied to downstream only, closing this PR.

pbrezina:
Please, keep this opened. There is a patch currently in Fedora only to fix this, but unfortunately, it causes some timeouts in resolved. See: https://bugzilla.redhat.com/show_bug.cgi?id=2291062 But eventually, it will get fixed.

pbrezina:
Well, it does not look like it is going to be fixed any time soon. Let's close this for now.

PR #386, opened 2025-01-13
sssd: pam_usertype already present in stack earlier

jcpunk:
This appears to remove a duplicate line in the pam stack.  Since `pam_usertype` is already present on line 7 having it again feels odd.


pbrezina:
What precisely feels odd about it? Did you experience any issue?

The first check is to skip localauth check for system users, so system users go always via pam_unix. The second check is the opposite, only regular users go through pam_sss.

jcpunk:
It appears I misunderstood the stack layout

PR #388, opened 2025-01-16
configure: fix typo

debarshiray:
Otherwise, it leads to:
```ShellSession
  $ ./configure
  ...
  checking where the gettext function comes from... libc
  ./configure: line 15565: ++ac_func=: command not found
  checking for reallocarray... yes
  ...
```

Fallout from b415003c3868d8b343d8bef4bce804571805e616


debarshiray:
Thanks, @pbrezina !

PR #389, opened 2025-01-24
profiles: add without-lastlog-showfailed

pbrezina:
Resolves: https://github.com/authselect/authselect/issues/354


thalman (approved):
ACK

traylenator:
Looks good - totally makes sense to be a new option rather than overload the existing `with-silent-lastlog` one.

Ta very much.

PR #398, opened 2025-05-05
profiles: replaces pam_lastlog with pam_lastlog2

pbrezina:
Resolves: https://github.com/authselect/authselect/issues/374


zhanhb:
`{if not "without-lastlog-showfailed":showfailed}` is removed again.

PR #421, opened 2025-09-17
Move apply-changes from scriptlet to systemd service

pbrezina:
To make it work correctly on ostree systems.

See:
* https://bugzilla.redhat.com/show_bug.cgi?id=2396016
* https://discussion.fedoraproject.org/t/pam-lastlog-change-crashing-gdm/164467

### DIFF
--- a/src/lib/Makefile.am
+++ b/src/lib/Makefile.am
@@ -86,7 +86,7 @@ libauthselect_la_CFLAGS = \
     $(NULL)
 libauthselect_la_LDFLAGS = \
     -Wl,--version-script=$(srcdir)/authselect.exports \
-    -version-info 5:0:1
+    -version-info 6:0:1
 
 pkgconfigdir = $(libdir)/pkgconfig
 pkgconfig_DATA = \


pbrezina:
Yeah, I will need to add preset. This PR is not yet finished though, I need to make the service run only if authselect profiles were updated. Unfortunately, ostree makes this difficult as /usr timestamps are set to 0 so I need to workaroud that. I'm working on it.

src/conf_macros.m4
```diff
@@ -23,12 +23,8 @@ CONFIGURABLE_VALUE(config-dir, config_dir, AUTHSELECT_CONFIG_DIR, DIR,
                    $sysconfdir/authselect)
 
 CONFIGURABLE_VALUE(profile-dir, profile_dir, AUTHSELECT_PROFILE_DIR, DIR,
-                   [Path to the directory where are stored authselect default profiles],
-                   $datarootdir/authselect/default)
-
-CONFIGURABLE_VALUE(vendor-dir, vendor_dir, AUTHSELECT_VENDOR_DIR, DIR,
-                   [Path to the directory where are stored profiles created by 3rd-party vendors],
-                   $datarootdir/authselect/vendor)
+                   [Path to the directory where are stored authselect default and vendor profiles],
```
thalman, line 30: A bit better: Path to the directory where authselect default and vendor profiles are stored

thalman (changes requested):
Minor string improvement suggestions.

systemd/authselect-apply-changes.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Apply authselect changes
+Documentation=man:authselect(8)
+After=local-fs.target
+Before=systemd-user-sessions.service nss-user-lookup.target
+
+[Service]
+Type=oneshot
+ExecStart=/usr/bin/authselect apply-changes --upgrade
+RemainAfterExit=yes
```
sgallagher, line 10: Is there a local file that will be present if this upgrade is run that wouldn't have been present prior to it?

If so, you can include a `ConditionPathExists=` or `ConditionPathIsSymlink=` line which will avoid running this command at all.

pbrezina: No, but I think I can use "ExecCondition=" and compare /var/lib/authselect/checksum with /usr/share/authselect/checksum.

pbrezina: Which is the same thing that `apply-changes --upgrade` do.

travier: Maybe we can use `ConditionNeedsUpdate=|/etc`. It should work on rpm-ostree systems.

pbrezina: It actually does not since /usr is mounted with 0 timestamps.

pbrezina: And how does it work? This is what I have on my bootc image

```
$ stat /usr
  File: /usr
  Size: 224       	Blocks: 8          IO Block: 4096   directory
Device: 0,39	Inode: 603         Links: 12
Access: (0755/drwxr-xr-x)  Uid: (    0/    root)   Gid: (    0/    root)
Context: system_u:object_r:usr_t:s0
Access: 1970-01-01 01:00:00.000000000 +0100
Modify: 1970-01-01 01:00:00.000000000 +0100
Change: 1970-01-01 01:00:00.000000000 +0100
 Birth: -

$ cat /etc/.updated 
# This file was created by systemd-update-done. Its only 
# purpose is to hold a timestamp of the time this directory
# was updated. See man:systemd-update-done.service(8).
TIMESTAMP_NSEC=0
```

```
directory requires an update because /usr/'s modification time is newer than the stamp file .updated in the specified directory
```

This is not true.

PR #425, opened 2025-11-05
Add new switchable-auth PAM stack

pbrezina:


profiles/sssd/switchable-auth
```diff
@@ -0,0 +1,44 @@
...
+auth      required                                     pam_deny.so
+
+account   required                                     pam_access.so                                          {include if "with-pamaccess"}
+account   required                                     pam_faillock.so                                        {include if "with-faillock"}
+account   sufficient                                   pam_systemd_home.so                                    {include if "with-systemd-homed"}
+account   required                                     pam_unix.so
+account   sufficient                                   pam_localuser.so                                       {exclude if "with-files-access-provider"}
+account   sufficient                                   pam_usertype.so issystem
+account   [default=bad success=ok user_unknown=ignore] pam_sss.so
+account   required                                     pam_permit.so
```
ikerexxe: Wouldn't it be better to include `password-auth`?

profiles/sssd/switchable-auth
```diff
@@ -0,0 +1,44 @@
...
+
+password  sufficient                                   pam_systemd_home.so                                    {include if "with-systemd-homed"}
+password  requisite                                    pam_pwquality.so local_users_only
+password  [default=1 ignore=ignore success=ok]         pam_localuser.so                                       {include if "with-pwhistory"}
+password  requisite                                    pam_pwhistory.so use_authtok                           {include if "with-pwhistory"}
+password  sufficient                                   pam_unix.so yescrypt shadow {if not "without-nullok":nullok} use_authtok
+password  [success=1 default=ignore]                   pam_localuser.so
+password  sufficient                                   pam_sss.so use_authtok
+password  required                                     pam_deny.so
```
ikerexxe: Same here

ikerexxe (commented):
Commits 2 and 3 mention `Resolves: https://github.com/authselect/`, but that's not correct. Apart from that I left some comments inline.

joantolo:
Built and tested locally. I've modified `gdm-switchable-auth` PAM service to add `switchable-auth` generated from here and everything worked fine. 👍

profiles/sssd/README
```diff
@@ -122,6 +122,17 @@ with-libvirt::
...
+with-switchable-auth::
+    Generate switchable-auth PAM stack that can be used by login applications
+    to select the authentication method that shall be used to authenticate the
+    user.
+
+with-switchable-web-auth::
+    Allow to select web-based authentication if available.
+
+with-switchable-passkey-auth::
+    Allow to select passkey authentication if available.
```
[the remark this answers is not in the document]
ikerexxe: > > Perhaps the default should be true for these new options?
> 
> I support this.

I think this is the way forward. Let's enable by default and the admin can decide to disable if needed.

profiles/sssd/switchable-auth
ikerexxe: I installed authselect latest update, modified `gdm-switchable-auth` to include `switchable-auth` and enabled the following `authselect` configuration:
```
Profile ID: sssd
Enabled features:
- with-sudo
- with-smartcard
- with-switchable-auth
- with-switchable-passkey-auth
- with-switchable-web-auth
- with-systemd-homed
```

All authentication mechanisms work and the user is able to change their password on expiration.

pbrezina: Ok, I updated it and included system-auth for account, session and password. It is the same as password-auth.

ikerexxe (approved):
LGTM! Thank you for putting this together.

joantolo (changes requested):

profiles/local/switchable-auth
```diff
@@ -0,0 +1 @@
+auth required pam_debug.so auth=authinfo_unavail
```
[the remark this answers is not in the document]
pbrezina: It is for sssd https://github.com/authselect/authselect/pull/425/commits/3444ffc498a730f6bf943d13092479277ad95954#diff-0361612e3581da745de2cb12631d211904fff5e4b31c77dc5d99ae8cbcc22896R7

If I get it correctly, you want to set it to false on other profiles as it does not make sense to use it there, right?

PR #432, opened 2026-01-16
backup-restore: ignore missing switchable-auth in backup

pbrezina:
This is a new file and it can be missing from the backup. In this case
we will just write an empty file. This is a corner case, an empty file
will not break system authentication so this is fine.

https://bodhi.fedoraproject.org/updates/FEDORA-2026-8a3906ef4b#comment-4516638

### DIFF
--- a/src/lib/authselect_backup.c
+++ b/src/lib/authselect_backup.c
@@ -240,17 +240,17 @@ static errno_t
 authselect_restore_system_configuration(const char *path)
 {
     struct selinux_safe_copy table[] = {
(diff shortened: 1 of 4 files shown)


pbrezina:
> I have one concern, the new file creation doesn't set any SELinux label, thus it will inherit the one from the parent. Is that fine?

In this case it should use the default context for the path. But anyway, I rewrote it to use `template_write_temporary` which also adds the authselect preambule to the file and also handles selinux. Please review it again.

PR #440, opened 2026-05-09
profiles: add support for oddjob-gpupdate to SSSD

Juff-Ma:
Samba for a while now included support for applying basic GPOs (sudo rules, banner, files, browser policies, etc.) to Linux machines via `samba-gpupdate`.

Usually applying them automatically is only supported when using winbind as the auth mechanism since it includes this built in. Alt linux, which includes their own gpo tool, has created `oddjob-gpupdate` in order to allow this functionality to be used with SSSD instead.

Originally OpenSUSE forked `oddjob-gpupdate` and made it compatible with the `samba-gpupdate` mechanism. This package is now also included in Fedora. Yet there has been no easy way to configure its PAM module since authselect does not include it in its SSSD profile.

This adds support for the `with-gpupdate` feature which relies on this module to allow group policies to be automatically applied when using SSSD. 

I placed the PAM module after SSSD since AFAIK samba-gpupdate should run as late as possible but before the GNOME keyring (if included) since samba includes GPOs that might change GNOME settings. However if anybody thinks there is a more logical order I'm obviously willing to make changes.

### DIFF
--- a/profiles/sssd/README
+++ b/profiles/sssd/README
@@ -132,6 +132,11 @@ with-switchable-auth::
     to select the authentication method that shall be used to authenticate the
     user.
 
+with-gpupdate::
+    Enable automatic application of Samba Group Policy Objects (GPOs) on user
+    login through oddjob-gpupdate. This allows SSSD-based systems to apply
+    GPOs (sudo rules, banners, files, browser policies) similar to winbind.
+
 EXAMPLES
 --------
 
--- a/profiles/sssd/REQUIREMENTS
+++ b/profiles/sssd/REQUIREMENTS
@@ -31,3 +31,7 @@ Make sure that SSSD service is configured and enabled. See SSSD documentation fo
                                                                                           {include if "with-switchable-auth"}
 - with-switchable-auth is selected, make sure to enable it in sssd.conf                   {include if "with-switchable-auth"}
   - set "pam_json_services = list-of-services" in [pam] section                           {include if "with-switchable-auth"}
+                                                                                          {include if "with-gpupdate"}
+- with-gpupdate is selected, make sure pam_oddjob_gpupdate module                         {include if "with-gpupdate"}
+  and samba-gpupdate are present and oddjobd service is enabled and active                {include if "with-gpupdate"}
+  - systemctl enable --now oddjobd.service                                                {include if "with-gpupdate"}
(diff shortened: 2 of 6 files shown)


pbrezina:
Hi, what does the tool do? Doesn't SSSD handle GPOs itself during access check?

Juff-Ma:
> Hi, what does the tool do? Doesn't SSSD handle GPOs itself during access check?

That is correct. SSSD implements GPOs for access limitations (and I think automounting as well).

But that's also the limit of what it supports. Samba's GPO support goes further than that. It can, for example, configure GNOME, manage the firewall, apply browser profiles (Chrome/Firefox) or distribute files. Many of these policies are also 1:1 compatible with their Windows equivalents (for example if configuring Chrome policies, they will work on both Linux and Windows seamlessly).

Samba's GPO support is distributed through the `samba-gpupdate` package. When using winbind it is invoked automatically. SSSD however needs the oddjob module.

Samba has a great [wiki article](https://wiki.samba.org/index.php/Group_Policy) about it and there's even an [online book](https://dmulder.github.io/group-policy-book/intro.html) going into further detail.

profiles/sssd/README
```diff
@@ -127,6 +127,9 @@ with-switchable-auth::
     to select the authentication method that shall be used to authenticate the
     user.
 
+with-gpupdate::
+    Apply samba group policies on user login.
```
pbrezina: How about:

```
Enable automatic application of Samba Group Policy Objects (GPOs) on user                                                                                                                                                               
login through oddjob-gpupdate. This allows SSSD-based systems to apply                                                                                                                                                                  
GPOs (sudo rules, banners, files, browser policies) similar to winbind.                                                                                                                                                                 
```

pbrezina (changes requested):
Hi, thank you for the explanation. This looks reasonable and I also talked to Pavel Filipensky who maintains it in Fedora. However, I'd like to improve the description in readme.

profiles/sssd/README
Juff-Ma: Sounds good. I made the change.

pbrezina (approved):
Thank you. I just squashed the two commits, let's wait for CI.

Juff-Ma:
Thank you very much!

PR #444, opened 2026-05-20
profiles: add with-group-merging feature

pbrezina:
This also makes group merging disabled by default. Group mergin
is not required by majority of users, but it makes all group lookups
to reach all providers (files, sss, systemd) which may cause performance
issues under some conditions.

It is better to let users who need it to enable it explicitly.


thalman (approved):
Code looks good, lets see what CI says.

PR #446, opened 2026-05-27
cli: clarify authselect check success messages

### LINKED ISSUE #437: Enhance output of `authselect check`
From `authselect(8)`:

>    How can I tell if my system is using authselect?
       Use authselect check. The output will tell you if you have 1) configuration generated by authselect 2) non-authselect configuration or 3)
       configuration that was generated by authselect but modified manually at some point.

**Actual behavior:**

```
]$ authselect check
Current configuration is valid.
$ echo $?
0
```

With this output I only know that my current configuration is valid. It does not provide any information about whether my config has been generated by `authselect` (1) or not (2). As I currently don't have any system using `authselect` I didn't check if the output looks different when using it.

IMHO the output can be improved as follows.

**Expected behavior (example):**

In case of 1):

```
]$ authselect check
Current authselect configuration is valid.
$ echo $?
0
```

In case of 2):

```
]$ authselect check
Current non-authselect configuration is valid.
$ echo $?
0
```

Tronde:
Distinguish valid authselect and non-authselect configuration in check output. Document the messages in authselect(8).

I did my best putting this together but still review it carefully.

Fixes: #437


src/cli/main.c
```diff
@@ -365,14 +365,14 @@ static errno_t check(struct cli_cmdline *cmdline)
...
-        puts(_("Current configuration is valid."));
+        puts(_("Current authselect configuration is valid."));
         break;
     case ENOENT:
         puts(_("No configuration detected."));
         ret = ENODEV;
         break;
     case EEXIST:
-        puts(_("System was not configured with authselect."));
+        puts(_("Current non-authselect configuration is valid."));
```
pbrezina: I wouldn't touch this. Authselect has no idea if non-authselect configuration is valid or not.

Tronde: Understood. I'll revert that and check the rest for alignment. Anything else I could improve or shouldn't have touched from your perspective?

Tronde: @pbrezina I revered that change and checked the rest for alignment. I'd appreciate it, if you find the time to take another look at it.

src/man/authselect.8.adoc
```diff
@@ -114,9 +114,20 @@ To print help for the selected command run *authselect COMMAND --help*.
...
 *check*::
-    Check if the current configuration is valid (it was either created by
-    *authselect* or there are no leftovers from previous authselect
-    configuration).
+    Check the current configuration.
+
+    On success, the command prints one of:
+
+    * *Current authselect configuration is valid.* -- configuration was
+      generated by *authselect* and was not modified manually.
```
pbrezina: Latest authselect version does not check for manual changes to the configuration, so I would strip this part and just say "configuration was generated by authselect".

src/man/authselect.8.adoc
```diff
@@ -114,9 +114,20 @@ To print help for the selected command run *authselect COMMAND --help*.
...
+    * *System was not configured with authselect.* -- the system is not
+      using *authselect* and there are no leftovers from a previous
+      *authselect* configuration.
+    * *No configuration detected.* -- no PAM or nsswitch configuration was
```
pbrezina: no PAM or nsswitch configration that is relevent to authselect...

(there might be other pam stacks in /etc/pam.d)

pbrezina (changes requested):
Hi, thank you. See two more comments and also please squash all commits into a single one. I think we can merge then.

Tronde:
Hi @pbrezina,
I did what you requested. Please give it another review if you find the time.

Cheers,  
Joerg

src/man/authselect.8.adoc
```diff
@@ -114,9 +114,19 @@ To print help for the selected command run *authselect COMMAND --help*.
...
+
+    * *Current authselect configuration is valid.* -- configuration was
+      generated by *authselect*.
+    * *System was not configured with authselect.* -- the system is not
+      using *authselect*.
+    * *No configuration detected.* -- no PAM or nsswitch configuration
+      relevant to *authselect*.
+      (there might be other pam stacks in /etc/pam.d)
+
+    If validation fails, the command prints an error and exits with code 3.
```
pbrezina: I would just refer to RETURN CODES section here so we do not duplicate the information.

src/man/authselect.8.adoc
```diff
@@ -270,8 +280,9 @@ TROUBLESHOOTING
 How can I tell if my system is using authselect?
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 Use *authselect check*. The output will tell you if you have 1) configuration
-generated by authselect 2) non-authselect configuration or 3) configuration
-that was generated by authselect but modified manually at some point.
+generated by authselect (*Current authselect configuration is valid.*) 2)
+non-authselect configuration (*System was not configured with authselect.*) or
+3) configuration that fails validation (error message, exit code 3).
```
pbrezina: Do you think this is helpful? It is already mentioned in the "check" command. I would rather choose only one place so we do not duplicate the information - maybe keep "check" command description short and put more structured information here?

pbrezina (changes requested):
Hi, I think the presented information is ok now. But see my small suggestions.

src/man/authselect.8.adoc
Tronde: I agree that it has benefits to avoid duplicate information. But I would prefer keeping the detailed information in the `check` command section. I might would like to understand what it does and returns without the particular question in mind: "How can I tell if my system is using authselect?"

I would change this section to:

~~~
How can I tell if my system is using authselect?                 
         
Use *authselect check*. See description of the `check` subcommand
for more information.
~~~

What do you think?

src/man/authselect.8.adoc
Tronde: I edited this to the following:

~~~
-    If validation fails, the command prints an error and exits with code 3.
+    If validation fails, the command prints an error and exits. See section
+    RETURN CODES for more information.
~~~

Is that what you have in mind?

src/man/authselect.8.adoc
pbrezina: Sounds good to me. I'm fine either way.

src/man/authselect.8.adoc
pbrezina: How about:

If the configuration is not valid, the command prints found errors.

You can refer to RETURN CODES section for specific error codes returned by the command.

src/man/authselect.8.adoc
Tronde: OK. Allow me some time to do the changes as I'm traveling the next four days.

pbrezina:
When you will be updating the commit, please, also add "Resolves: https://github.com/authselect/authselect/issues/437" to the commit message.

Tronde:
> When you will be updating the commit, please, also add "Resolves: #437" to the commit message.

Thank you for your patience and advice. I included the requested changes in the latest commit.

Cheers,  
Joerg

pbrezina:
Thank you, ack. I just added link to the issue to the commit message, let's wait for PR CI and merge.

PR #458, opened 2026-06-21
sssd: Ignore unknown users in pam_sss for with-smartcard-required

### LINKED ISSUE #457: sssd: ignore unknown users in pam_sss when with-smartcard-required is set
This issue is related to: #305 

The bug was fixed for `with-smartcard` but not `with-smartcard-required` [BZ 2070325](https://bugzilla.redhat.com/show_bug.cgi?id=2070325) 

Impact: EL9 and EL8

```
# rpm -q authselect-libs sssd-client
authselect-libs-1.2.6-3.el9.x86_64
sssd-client-2.9.8-4.el9_8.x86_64
```

asakure:
This patch ensures to ignore unknown users in pam_sss when with-smartcard-required is set.

Resolves: https://github.com/authselect/authselect/issues/457


pbrezina:
Hi, thank you for the contribution, but I must decline it.

The intention behind smartcard-required is to require smartcard for the whole system. In this case, even local users must be configured to authenticate with smartcards via SSSD, if not, it is considered an error and the authentication will be denied. Thus the current state is intentional.

PR #460, opened 2026-06-23
tests: migrate system tests to authselect-test-framework

danlavu:
Switch the system test suite from sssd-test-framework to
authselect-test-framework. https://github.com/danlavu/authselect-test-framework

- Update conftest, requirements, and mhc.yaml for Local, SSSD, and Winbind profiles
- Drop unused NFS and AD CI setup from mhc.yaml and GitHub Actions workflow
- Rework local, SSSD, and winbind feature tests with inline docstrings
- Use authselect-first flows and assert_selected() after every profile selection
- Add winbind system tests and refresh CLI and preset coverage
- Add ruff lint configuration for the system test tree


pbrezina:
Can you instead of reverting the commit simply delete the commit from here?

danlavu:
Done.

PR #463, opened 2026-07-30
Update fingerprint authentication condition in dconf-db

### LINKED ISSUE #462: `with-systemd-homed` should disable GDM fingerprint auth
Currently the dconf overlay to GDM reads thus:

```
[org/gnome/login-screen]
enable-smartcard-authentication=false
enable-fingerprint-authentication={if "with-fingerprint":true|false}
enable-switchable-authentication=false
```

However, letting GDM allow logins with the fingerprint scanner prevents the password from traversing the PAM login flow and fails to decrypt the user's home. Encrypted-home setups like `systemd-homed` or `ecryptfs` should cause the `enable-fingerprint-authentication` flag to be set to false, to continue allowing fingerprint _unlocks_ but disallow fingerprint _logins_.

Thesola10:
This fixes #462 by ensuring GDM will not attempt to unlock the session with a login pathway that fails to set `PAM_AUTHTOK` if the user's home requires it for decryption.

### DIFF
--- a/profiles/local/dconf-db
+++ b/profiles/local/dconf-db
@@ -1,4 +1,4 @@
 [org/gnome/login-screen]
 enable-smartcard-authentication=false
-enable-fingerprint-authentication={if "with-fingerprint":true|false}
+enable-fingerprint-authentication={if "with-fingerprint" and not ("with-systemd-homed" or "with-ecryptfs"):true|false}
 enable-switchable-authentication=false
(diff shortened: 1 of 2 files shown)


joantolo:
Since GNOME 51, GDM doesn't allow authentication with fingerprint on the login screen, but it's allowed on the lockscreen.

So this change shouldn't be needed for distros with GNOME 51 (Ubuntu 26.10, Fedora 45...) 

And I think if you disable fingerprint with authselect this way, it's disabled by default also for the lockscreen.

I guess if you want this for systems that don't use GNOME 51, you could print some info mentioning how to enable fingerprint authentication for a user session when fingerprint gets disabled because of systemd-homed or ecryptfs.

Thesola10:
The best course of action would just be to wait for Fedora 45 then?

Since I'm not sure where discussion on this project happens, I'd like to ask here: is a tag planned soon? #429 was merged a while ago but still hasn't been released.

pbrezina:
I will make a new release for F45 reasonably soon.

Thank you Joan, let's wait for Gnome 51 for this.

